Issue #4, opened 2015-03-23
npm prune option


remixz:
would be nice to have an option that runs `npm prune --production` on the app, then people could add this and `atom-shell` to dev deps, and it just removes them

maxogden:
great idea, it could even run it after it copied the app into the App folder so its non-destructive to the source folder

it might be worth using https://github.com/datproject/gasket for automating the multi-process step stuff if this gets more complex, e.g. https://github.com/meandavejustice/gh-pages-deploy/blob/master/index.js

PR #7, opened 2015-03-23
Add prune option

remixz:
Fixes GH-4.

When run with `--prune`, it'll run on `npm prune --production` on the app folder. Leaves the original directory's modules in tact.


maxogden:
@remixz kewl thanks. I did some slight code reorganizing in https://github.com/maxogden/atom-shell-packager/commit/cbc585393bdc256f509ba1070c0ed42662e98f51 (hoisting and separation of concerns)

PR #9, opened 2015-04-04
bump atom-shell

mafintosh:


freeall:
:+1:

maxogden:
released 2.0.0 which uses require.resolve to find atom-shell instead of downloading a new one. this is also nicer IMO as atom-shell is a kind of big download

Issue #2, opened 2015-03-23
add asar support


maxogden:
maybe via an `asar: true` option that would simply run `asar pack` for you

PR #18, opened 2015-05-06
Add asar support. Closes GH-2

shama:
`electron-packager . MyApp --asar`

Thanks!

### DIFF
--- a/linux.js
+++ b/linux.js
@@ -3,6 +3,8 @@ var fs = require('fs')
 var child = require('child_process')
 var mkdirp = require('mkdirp')
 var ncp = require('ncp').ncp
+var rimraf = require('rimraf')
+var asar = require('asar')
 
 module.exports = {
   createApp: function createApp (opts, cb, electronPath) {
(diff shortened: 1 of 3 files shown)


linux.js
```diff
@@ -43,6 +45,11 @@ module.exports = {
     function renameElectronBinary () {
       fs.rename(originalBinary, finalBinary, function electronRenamed (err) {
         if (err) return cb(err)
+        if (opts.asar) {
+          asarApp(cb)
+        } else {
+          cb()
```
malept: Is this necessary?

shama: I thought so as it didn't seem to call `cb()` which for the cli probably doesn't matter but if someone was using it as a lib they would want to know when the `cb()` was called.

mafintosh:
@malept like @shama said we need to call the callback for module support

mafintosh:
Hmm it seems i don't have publish rights. I already bumped the version so as soon as @maxogden wakes up I can publish this to npm.

maxogden:
published

PR #26, opened 2015-05-11
Validate the --arch argument

Ivshti:
Currently electron-packager tries to download the electron archive with the given arch, no matter if the arch is valid or not.


maxogden:
fixed what @jden mentioned and published as 4.0.2

Ivshti:
Awesome, thanks.

Issue #16, opened 2015-04-26
sign app using xcode


maxogden:
as seen here https://github.com/Aluxian/Facebook-Messenger-Desktop/blob/master/gulpfile.coffee

kfranqueiro:
Should this be closed given that `--sign` is implemented?

malept:
Looks like it.

PR #31, opened 2015-05-12
Added --sign to mac

jenslind:
It's a start. Starting to get ugly with all them callbacks though. :)

https://github.com/maxogden/electron-packager/issues/16

### DIFF
--- a/mac.js
+++ b/mac.js
@@ -108,24 +108,30 @@ function buildMacApp (opts, cb, newApp) {
           if (opts.asar) {
             asarApp(function (err) {
               if (err) return cb(err)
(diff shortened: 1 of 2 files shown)


maxogden:
nice. what else does this need beyond what you have now?

jenslind:
Some input, I'm not too familiar with `codesign`. Also wondering if there should be a verification of the sign afterwards but i guess its not needed because it should throw an error if it fails anyway.

But yeah, what's in the commit works.

maxogden:
gonna merge for now

maxogden:
published as 4.1.0

Issue #13, opened 2015-04-23
Ignore `electron-prebuilt` and `electron-packager` by default


sindresorhus:
So I don't have to `$ atom-shell-packager . App --ignore=node_modules/electron-prebuilt --ignore=node_modules/electron-packager`.

remixz:
If you have those dependencies in your `devDependencies`, you can pass `--prune`, and it'll remove those from the bundle, along with any other `devDependencies`.

In any case though, :+1: to that, would be a good default.

maxogden:
patch welcome :)

sindresorhus:
Should also exclude the `.git` folder by default.

stefanbuck:
Maybe should get the ignores from the `.gitignore`

sindresorhus:
Or like npm, either use the whitelisting `files` array in package.json or the .npmignore blacklist.

jden:
:+1: for copying the npm ignore algorithm:

whitelist `files` in package.json || blacklist .npmignore || blacklist .gitignore || *

englercj:
:+1: For using the npm algorithm but I don't think that is enough to fix the need I have with choosing the right node_modules. I won't put "node_modules/..." in my package.json files array, because I wouldn't want them in my npm tgz, but there are like 2 modules that need to be included in the electron package.

I have a bunch of normal dependencies that are used at runtime, but are compiled into the renderer code (browserify). I only have like 1 or 2 modules of my "dependencies" that I want to include in the output package that are used by the app code. I need a way to *whitelist* modules instead of only allowing *blacklisting*.

jenslind:
This would be great! Could the use of a custom field like `buildFiles` instead of `files` be an option? Just to not interfere with npm.

maxogden:
this seems like a problem not specific to this module, so a generic solution would be preferred, e.g. the ability for the user to run a pre-package script that did whatever they want, and then someone can make a 'strip-non-essential-files' CLI module that can be used there

jden:
@sindresorhus is there anything else in this issue that was not addressed by #24? Can it be closed?

sindresorhus:
@maxogden Not optimal as the user probably don't want those files deleted, only ignored from the packaged app, and the script would have no knowledge of how `electron-packager` packages the apps. The ignore logic could of course be a separate module `electron-packager` depends on.

@jden When we have an agreement on the above.

malept:
[...]

PR #34, opened 2015-05-18
Ignore this and related modules

jenslind:
Regardless of what the outcome in https://github.com/maxogden/electron-packager/issues/13 will be. I think we should ignore what @sindresorhus first suggested, the module itself and the related `electron-prebuilt` and even `.git`.

:construction_worker:

### DIFF
--- a/index.js
+++ b/index.js
@@ -31,6 +31,11 @@ module.exports = function packager (opts, cb) {
     default: return cb(new Error('Unsupported platform. Must be either darwin, linux, or win32'))
   }
 
+  // Ignore this and related modules by default
+  var default_ignores = ['node_modules/electron-prebuilt', 'node_modules/electron-packager', '.git']
+  if (opts.ignore && !Array.isArray(opts.ignore)) opts.ignore = [opts.ignore]
+  opts.ignore = (opts.ignore) ? opts.ignore.concat(default_ignores) : default_ignores
+
   download({
     platform: platform,
     arch: arch,


sindresorhus:
:+1:

sbruchmann:
Although [electron-gh-releases](https://github.com/jenslind/electron-gh-releases) is quite unpopular compared to _electron-prebuilt_ and _electron-packager_, it is a related module. Would it make sense to ignore that too?

jenslind:
@sbruchmann no, `electron-gh-releases` is not a dev-dependencie and therefor it's needed in the packaged app.

sbruchmann:
@jenslind Ah, yes. I’ve missed that one.

maxogden:
+1, though I use camelCase in the rest of the code .I'll merge and fix

maxogden:
published as 4.0.3

jden:
:+1: all around :ocean:

jits:
Is there a way to explicitly _include_ `electron-prebuilt` as a dependency in my built app? I'm using `child_process` to spawn worker processes for data processing, by calling on the `electron` executable (via `electron = require('electron-prebuilt')`). Which fails in my built app.

Alternatively is there a better way to spawn child processes that run a simple NodeJS script?

malept:
@jits have you looked at https://github.com/electron/electron/issues/797 ?

jits:
@malept – thanks for the suggestion. I have considered Web Workers, but was having trouble getting any decent results (running 10K+ jobs) with them (possibly due to my lack of awareness of how it all fits). Also, I'm relying on some node modules that may not easily run in the browser, and using `electron.remote` is not possible in workers (and could be a major performance blocker anyway).

As for IPC'ing + spawning BrowserWindows – another reasonable option, but also may not be performant enough. More importantly, I want to avoid writing my own job management library and instead rely on something that works within the community. So, using https://github.com/bjrmatos/electron-workers works really well for my use case… except for the built app. [Under the hood](https://github.com/bjrmatos/electron-workers/blob/2e51443d3d1539a08f29c7077d903c69dc9f1d6b/src/ElectronManager.js#L28-L40) it does `electron = require('electron-prebuilt')` or looks for a system level `electron` (is it possible to install that?)

Any more suggestions to allow my app to call on an electron executable are greatly appreciated!

malept:
@jits you can bypass the default ignores by setting the [`ignore` parameter](https://github.com/electron-userland/electron-packager/blob/master/docs/api.md#ignore) to be a `Function`. Obviously, you cannot do this in the CLI - it is only available if you use the API.

jits:
Thanks @malept – good to know. Alas I'm currently using the CLI. May wrap this in a gulp task or something.

PR #42, opened 2015-05-20
symlink dereference added

Rameshv:
Hi, 
  This pull request adds a new option `dereference` which is passed to the `options.dereference` in https://github.com/AvianFlu/ncp. This allows the build to copy symlinked resources as regular resources to the final directory.

This is very helpful when you have your `node_modules` and `bower_components` are symlinked from the other directories. 

Default value is `false` as `ncp`

### DIFF
--- a/linux.js
+++ b/linux.js
@@ -29,7 +29,7 @@ module.exports = {
 
     function copyUserApp () {
       var ncpOpts = {
-        dereference : opts.dereference || false,
+        dereference: opts.dereference || false,
         filter: userFilter
       }
       ncp(opts.dir, userAppDir, ncpOpts, function copied (err) {
--- a/linux.js
+++ b/linux.js
@@ -28,7 +28,11 @@ module.exports = {
     }
 
     function copyUserApp () {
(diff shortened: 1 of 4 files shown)


jden:
Great idea! Please check that this passes `npm test` and then [rebase against master](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html)

maxogden:
instead of specifically passing dereference we should just have a way to pass a custom ncp options object in, that way we can just support 100% of ncp options

jden:
Actually, what about just passing dereference by default? If you're packaging up an electron app, presumably you're going to distribute it somewhere, and that system probably doesn't care about your symlinks. Copying the files is the only logical thing you would want to do when packaging them.

As for exposing all of ncp options, I'm -1, because I don't think that implementation detail should be exposed.

maxogden:
@jden good point, +1

sindresorhus:
:+1: what @jden said.

Rameshv:
@jden thats a good point, i am up for having the deference as a default option.. so https://github.com/maxogden/electron-packager/pull/43 is a right way to go

Issue #44, opened 2015-05-21
default ignores and --ignore not working (windows)


zaggino:
Hi, I'm using latest packager `4.1.0` and building the electron app using (in msysgit bash shell):
```
npm run bundle-win
```
in package.json I've defined
```
"bundle-win": "bundle.sh win32 x64 0.26.0",
```
which executes following script:
```
#!/bin/sh
electron-packager . Bracketron --platform=$1 --arch=$2 --version=$3 --ignore=node_modules/electron-rebuild
```
but the problem is, that neither `electron-rebuild` or default excludes (https://github.com/maxogden/electron-packager/commit/892b15fc306ee67bd1ea8a6aa6a1c799443b37f2) are being excluded from the build folder (otherwise it builds fine)

repo is here https://github.com/zaggino/brackets-electron

jenslind:
Hmm, I can't reproduce this issue on Mac. Does your linux and mac builds work as expected?

zaggino:
Can you please try the repo above? https://github.com/zaggino/brackets-electron
`npm run bundle-mac` doesn't even copy the app files into the `Contents/Resources/app` folder for some reason (it does copy on Windows but ignores excludes)

zaggino:
Nevermind, found one issue, going to submit a PR

zaggino:
PR for Mac is up, I'll check what's wrong on Windows too

zaggino:
PR for Windows is up too.

zaggino:
@maxogden another issue with `4.1.1`:
I have this in my shell script:
```
electron-packager . Bracketron --platform=$1 --arch=$2 --version=$3 --ignore='/node_modules/electron-rebuild($|/)'
```
but minimist will convert this to:
```
C:/Program Files (x86)/Git/node_modules/electron-rebuild($|/)
```
I've put console.log here: https://github.com/maxogden/electron-packager/blob/master/cli.js#L4

zaggino:
Actually this is done even before it reaches minimist for some reason, sorry. I need to check why.

zaggino:
Ok, I actually need to do
```
--ignore='./node_modules/electron-rebuild$'
```
to make it work without resolving the first `/` to `C:/Program Files (x86)/Git/`

But it excludes correctly now. You can close this or make some notes based on my findings into the README if you wish. Thanks for merging the PRs.

zaggino:
[...]

PR #45, opened 2015-05-22
fix defaultIgnores to be proper reg-exps

zaggino:
this fixes my case mentioned in https://github.com/maxogden/electron-packager/issues/44 - i had my project in a folder `/Users/zaggino/github/brackets-electron/` and `.git` matched the `/github` part

### DIFF
--- a/index.js
+++ b/index.js
@@ -32,7 +32,7 @@ module.exports = function packager (opts, cb) {
   }
 
   // Ignore this and related modules by default
-  var defaultIgnores = ['node_modules/electron-prebuilt', 'node_modules/electron-packager', '.git']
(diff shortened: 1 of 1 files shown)


jden:
For the sake of documenting the rationale for this PR in this thread, could you describe where the filter is being processed and why regexes are necessary?

zaggino:
They're processed here: https://github.com/maxogden/electron-packager/blob/master/mac.js#L71-L80

`if (file.match(ignore[i])) {` with ignore being `.git` will match even folders like `/github` or `mygitapp`

zaggino:
btw README doesn't say that `--ignore` are processed as regular expressions, but according to code they are for all three platforms

PR #47, opened 2015-05-22
merge filter functions into one and make it project relative

zaggino:
For the reasons described in https://github.com/maxogden/electron-packager/issues/44

Before the change the ignore string were treated as regexp matched against full file paths

After the change the ignore strings are treated as regexps but matched against relative file paths

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -1,10 +1,11 @@
 #!/usr/bin/env node
 var fs = require('fs')
+var path = require('path')
 var args = require('minimist')(process.argv.slice(2), {boolean: ['prune', 'asar']})
 var packager = require('./')
 var usage = fs.readFileSync(__dirname + '/usage.txt').toString()
 
-args.dir = args._[0]
+args.dir = path.resolve(args._[0])
 args.name = args._[1]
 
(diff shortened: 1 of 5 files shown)


jden:
@zaggino thanks for your efforts in investigating & PRing  these issues, especially cross-platform. Can you think of any ways of meaningfully testing this to avoid introducing subtle breaking changes in the future?

zaggino:
Sure, we could put together some test basic hello world project in this repo and run assertions after it's packaged through node's fs.

But that'd test only the ignores, I'm not sure if everything else would be testable. I haven't really digged through other code than the ignore parts.

kfranqueiro:
Based on the description and what I'm seeing, it sounds like this PR is now redundant of a combination of refactoring that's already been merged and #55 (which ironically the refactoring clobbered).  I think the cleanup/fixes in #88 will cover this.

kfranqueiro:
@zaggino, can you have a try with 5.0.0 and confirm if this is already resolved there?  Thanks!

zaggino:
I don't work at the project I used to before at the time being, but it seems this can be closed.

Issue #54, opened 2015-06-01
ignore filter matching to path in cwd 


chentsulin:
nsp's filter test regexp matching to a whole path name

```
/Users/chentsulin/Projects/test/electron-react-boilerplate
```
will match `'/test($|/)'` in `ignore` option:

```js
ignore: [
    '/test($|/)'
]
```

It will ignore copy of whole app and cause some unexpected behaviors.

jden:
The strategy in #55 of not matching `--ignore` patterns against the cwd works, but it seems like it should ignore the `<sourcedir>` instead, which may or may not be the cwd

chentsulin:
Is it appropriate if I use `require('path').resolve(opts.dir)` instead of `process.cwd()`?

kfranqueiro:
This was never closed even though #55 was merged.  However, it has resurfaced (see #84).  I will be including a fix to reinstate this fix in the branch I am planning to PR shortly.

kfranqueiro:
Now that #88 is merged, this should again be resolved.  Any objections to closing this?

malept:
No objections, so closing.

PR #55, opened 2015-06-01
get rid of source dir in path for ignore filter matching

chentsulin:
#54


jden:
Please update the PR to fix the failing build & squash the commits :)

PR #61, opened 2015-06-10
Deduplicate code + cleanup

malept:
Moves a few common functions to their own module, plus some code cleanup I noticed while I was moving things around.


malept:
I fixed the naming convention problem - if there are no further objections, I'll merge this by this weekend.

maxogden:
lgtm!

PR #69, opened 2015-06-16
Add quotes around signing identity to allow spaces

mmuehlberger:
The packager strips quotes from the `--sign` argument, which makes using a more detailed signing identity (necesssary when having multiple identities for iPhone, Mac App Store and/or Deveoper ID) impossible.


jden:
published as electron-packager@4.1.3

PR #74, opened 2015-06-18
Fix renaming folders across volumes (when packaging for OS X)

### LINKED ISSUE #60: Error building from non-OS volume
If my Electron application is on a different volume than my OS `electron-packager ./ MyApp …` fails with `EXDEV: cross-device link not permitted` when it gets to https://github.com/maxogden/electron-packager/blob/master/mac.js#L106 and tries to move (instead of copy) the packaged application to the output directory.

malept:
I put `moveDirectory` in `common.js` because it didn't seem like a OS-specific function.

Fixes #60.

### DIFF
--- a/mac.js
+++ b/mac.js
@@ -6,6 +6,7 @@ var child = require('child_process')
 var plist = require('plist')
 var mkdirp = require('mkdirp')
 var rimraf = require('rimraf')
+var mv = require('mv')
 var ncp = require('ncp').ncp
 var common = require('./common')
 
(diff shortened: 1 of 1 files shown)


kfranqueiro:
I have not tested this yet and I'm not entirely clear on how the fallback works, but wouldn't it perform a copy rather than a move?  Should we be cleaning up after it?

malept:
It works the same as atom/scandal#13, which does not seem to have any cleanup code.

malept:
Now uses `mv` instead, per discussion in #78.

PR #80, opened 2015-06-20
Switching to the positive, only grabbing specified files

krainboltgreene:
We need to be able to specify the file globs we want in the application, so we're reading from the package.json config. Instead of copy dir/ignore, positively pass paths based on dir.

### DIFF
--- a/index.js
+++ b/index.js
@@ -15,8 +15,12 @@ module.exports = function packager (opts, cb) {
   var platform = opts.platform
   var arch = opts.arch
   var version = opts.version
+  var config = require(__dirname + '/package.json').config
 
-  if (!platform || !arch || !version) cb(new Error('Must specify platform, arch and version'))
+  if (!platform) cb(new Error('Please specify a platform with --platform=, see --help'))
+  if (!arch) cb(new Error('Please specify a arch with --arch=, see --help'))
+  if (!version) cb(new Error('Please specify an Electron version with --version=, see --help'))
+  if (!config) cb(new Error('Please specify a package.json for your app'))
 
   switch (arch) {
     case 'ia32': break
@@ -31,10 +35,11 @@ module.exports = function packager (opts, cb) {
     default: return cb(new Error('Unsupported platform. Must be either darwin, linux, or win32'))
   }
 
-  // Ignore this and related modules by default
-  var defaultIgnores = ['/node_modules/electron-prebuilt($|/)', '/node_modules/electron-packager($|/)', '/\.git($|/)']
-  if (opts.ignore && !Array.isArray(opts.ignore)) opts.ignore = [opts.ignore]
-  opts.ignore = (opts.ignore) ? opts.ignore.concat(defaultIgnores) : defaultIgnores
+  // Only include files specified
+  var defaultFiles = ['package.json']
+  opts.files = (defaultFiles + config.files).map(function (item) {
+    path.join(opts.dir, item)
+  })
 
   download({
     platform: platform,
--- a/common.js
+++ b/common.js
@@ -25,26 +25,5 @@ module.exports = {
     } else {
       nextStep()
     }
-  },
-
-  userIgnoreFilter: function userIgnoreFilter (opts, is_win32, finalDir) {
-    return function filter (file) {
-      if (is_win32) {
-        // convert slashes so unix-format ignores work
-        file = file.replace(/\\/g, '/')
-      }
-
-      var ignore = opts.ignore || []
-      if (!Array.isArray(ignore)) ignore = [ignore]
-      if (typeof finalDir !== 'undefined') {
-        ignore = ignore.concat([finalDir])
-      }
-      for (var i = 0; i < ignore.length; i++) {
-        if (file.match(ignore[i])) {
-          return false
-        }
-      }
-      return true
-    }
   }
 }
--- a/linux.js
+++ b/linux.js
@@ -26,7 +26,7 @@ module.exports = {
     }
 
     function copyUserApp () {
-      ncp(opts.dir, userAppDir, {filter: common.userIgnoreFilter(opts, false, finalDir), dereference: true}, function copied (err) {
+      ncp(opts.files, userAppDir, {dereference: true}, function copied (err) {
         if (err) return cb(err)
         common.prune(opts, userAppDir, cb, renameElectronBinary)
       })
--- a/mac.js
+++ b/mac.js
@@ -68,8 +68,8 @@ function buildMacApp (opts, cb, newApp) {
   fs.writeFileSync(paths.info1, plist.build(pl1))
   fs.writeFileSync(paths.info2, plist.build(pl2))
 
-  // copy users app into .app
-  ncp(opts.dir, paths.app, {filter: common.userIgnoreFilter(opts), dereference: true}, function copied (err) {
+  // copy users files into .app
+  ncp(opts.files, paths.app, {dereference: true}, function copied (err) {
     if (err) return cb(err)
 
     function moveApp () {
--- a/win32.js
+++ b/win32.js
@@ -49,8 +49,8 @@ function buildWinApp (opts, cb, newApp) {
     app: path.join(newApp, 'resources', 'app')
   }
 
-  // copy users app into destination path
-  ncp(opts.dir, paths.app, {filter: common.userIgnoreFilter(opts, true), dereference: true}, function copied (err) {
+  // copy users files into destination path
+  ncp(opts.files, paths.app, {dereference: true}, function copied (err) {
     if (err) return cb(err)
 
     function moveApp () {


krainboltgreene:
There are no tests so I'm not sure what I've broken yet. I'm going to try and use my own fork to test my application.

index.js
```diff
@@ -31,10 +35,11 @@ module.exports = function packager (opts, cb) {
...
   }
 
-  // Ignore this and related modules by default
-  var defaultIgnores = ['/node_modules/electron-prebuilt($|/)', '/node_modules/electron-packager($|/)', '/\.git($|/)']
-  if (opts.ignore && !Array.isArray(opts.ignore)) opts.ignore = [opts.ignore]
-  opts.ignore = (opts.ignore) ? opts.ignore.concat(defaultIgnores) : defaultIgnores
+  // Only include files specified
+  var defaultFiles = ['package.json']
+  opts.files = (defaultFiles + config.files).map(function (item) {
+    path.join(opts.dir, item)
```
maxogden: I think you need a `return` here

maxogden: also just to clarify, is the idea to change the CLI API from `electron-packager <sourcedir>` to e.g. `electron-packager --files=foo/bar.js --files=foo/baz.js` where you specify each individual file on the CLI?

krainboltgreene: Yeah, I'm finding a lot of errors just by running things in my env. I'll
update this PR as the day goes.

On Sat, Jun 20, 2015 at 12:11 PM, maxogden <notifications@github.com> wrote:

> In index.js
> <https://github.com/maxogden/electron-packager/pull/80#discussion_r32886107>
> :
>
> > @@ -31,10 +35,11 @@ module.exports = function packager (opts, cb) {
> >      default: return cb(new Error('Unsupported platform. Must be either darwin, linux, or win32'))
> >    }
> >
> > -  // Ignore this and related modules by default
> > -  var defaultIgnores = ['/node_modules/electron-prebuilt($|/)', '/node_modules/electron-packager($|/)', '/\.git($|/)']
> > -  if (opts.ignore && !Array.isArray(opts.ignore)) opts.ignore = [opts.ignore]
> > -  opts.ignore = (opts.ignore) ? opts.ignore.concat(defaultIgnores) : defaultIgnores
> > +  // Only include files specified
> > +  var defaultFiles = ['package.json']
> > +  opts.files = (defaultFiles + config.files).map(function (item) {
> > +    path.join(opts.dir, item)
>
> I think you need a return here
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/maxogden/electron-packager/pull/80/files#r32886107>.
>



-- 
Kurtis Rainbolt-Greene, Hacker
Software Developer
1631 8th St.
New Orleans, LA
70115

maxogden: Oh I see, you are reading the `files` field of the `package.json` and using that as the list of files to include. I like that approach but not all package.jsons will have the `files` field populated, so I think we should maybe show the user an error or warning if the packager can't find any files to include.

Also is there a way to specify the files using flags like in my comment above? From reading the code it looks like it only pulls from `package.json`

krainboltgreene: Yes, I ultimately want this PR to encompass that change.

On Sat, Jun 20, 2015 at 12:12 PM, maxogden <notifications@github.com> wrote:

> In index.js
> <https://github.com/maxogden/electron-packager/pull/80#discussion_r32886121>
> :
>
> > @@ -31,10 +35,11 @@ module.exports = function packager (opts, cb) {
> >      default: return cb(new Error('Unsupported platform. Must be either darwin, linux, or win32'))
> >    }
> >
> > -  // Ignore this and related modules by default
> > -  var defaultIgnores = ['/node_modules/electron-prebuilt($|/)', '/node_modules/electron-packager($|/)', '/\.git($|/)']
> > -  if (opts.ignore && !Array.isArray(opts.ignore)) opts.ignore = [opts.ignore]
> > -  opts.ignore = (opts.ignore) ? opts.ignore.concat(defaultIgnores) : defaultIgnores
> > +  // Only include files specified
> > +  var defaultFiles = ['package.json']
> > +  opts.files = (defaultFiles + config.files).map(function (item) {
> > +    path.join(opts.dir, item)
>
> also just to clarify, is the idea to change the CLI API from electron-packager
> <sourcedir> to e.g. electron-packager --files=foo/bar.js
> --files=foo/baz.js where you specify each individual file on the CLI?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/maxogden/electron-packager/pull/80/files#r32886121>.
>



-- 
Kurtis Rainbolt-Greene, Hacker
Software Developer
1631 8th St.
New Orleans, LA
70115

krainboltgreene: I'm currently working on that now.

On Sat, Jun 20, 2015 at 12:16 PM, maxogden <notifications@github.com> wrote:

> In index.js
> <https://github.com/maxogden/electron-packager/pull/80#discussion_r32886167>
> :
>
> > @@ -31,10 +35,11 @@ module.exports = function packager (opts, cb) {
> >      default: return cb(new Error('Unsupported platform. Must be either darwin, linux, or win32'))
> >    }
> >
> > -  // Ignore this and related modules by default
> > -  var defaultIgnores = ['/node_modules/electron-prebuilt($|/)', '/node_modules/electron-packager($|/)', '/\.git($|/)']
> > -  if (opts.ignore && !Array.isArray(opts.ignore)) opts.ignore = [opts.ignore]
> > -  opts.ignore = (opts.ignore) ? opts.ignore.concat(defaultIgnores) : defaultIgnores
> > +  // Only include files specified
> > +  var defaultFiles = ['package.json']
> > +  opts.files = (defaultFiles + config.files).map(function (item) {
> > +    path.join(opts.dir, item)
>
> Oh I see, you are reading the files field of the package.json and using
> that as the list of files to include. I like that approach but not all
> package.jsons will have the files field populated, so I think we should
> maybe show the user an error or warning if the packager can't find any
> files to include.
>
> Also is there a way to specify the files using flags like in my comment
> above? From reading the code it looks like it only pulls from package.json
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/maxogden/electron-packager/pull/80/files#r32886167>.
>



-- 
Kurtis Rainbolt-Greene, Hacker
Software Developer
1631 8th St.
New Orleans, LA
70115

maxogden: Also in general (sorry for the deluge of comments) I am open to the 'positive' approach taken here, but I'm not sure we should remove the old way of doing it (where you filter stuff out). I think before making any breaking changes it would be a good idea to open a discussion issue to get feedback from the other maintainers/users on what we think a good set of APIs would be.

krainboltgreene: :+1:

krainboltgreene:
Ugh, just ran into an issue where I discovered `node-ncp` doesn't take an Array for the source argument. Thoughts?

maxogden:
@krainboltgreene ahh good point... you might have to still use a filter, but make it a positive filter rather than a negative one. maybe we should just expose the filter to the user in a better way.

krainboltgreene:
Okay, so good point, one of the big reasons I'm doing this is because I'm
having issues with file names and regular expressions.

Maybe if we pass the files through a regexp escaper?

On Sat, Jun 20, 2015 at 1:09 PM, maxogden <notifications@github.com> wrote:

> @krainboltgreene <https://github.com/krainboltgreene> ahh good point...
> you might have to still use a filter, but make it a positive filter rather
> than a negative one. maybe we should just expose the filter to the user in
> a better way.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/maxogden/electron-packager/pull/80#issuecomment-113797821>
> .
>



-- 
Kurtis Rainbolt-Greene, Hacker
Software Developer
1631 8th St.
New Orleans, LA
70115

krainboltgreene:
I'm about to just rewrite this PR, so don't depend on anything here yet.

On Sat, Jun 20, 2015 at 1:12 PM, Kurtis Rainbolt-Greene <
me@kurtisrainboltgreene.name> wrote:

> Okay, so good point, one of the big reasons I'm doing this is because I'm
> having issues with file names and regular expressions.
>
> Maybe if we pass the files through a regexp escaper?
>
> On Sat, Jun 20, 2015 at 1:09 PM, maxogden <notifications@github.com>
> wrote:
>
>> @krainboltgreene <https://github.com/krainboltgreene> ahh good point...
>> you might have to still use a filter, but make it a positive filter rather
>> than a negative one. maybe we should just expose the filter to the user in
>> a better way.
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/maxogden/electron-packager/pull/80#issuecomment-113797821>
>> .
>>
>
>
>
> --
> Kurtis Rainbolt-Greene, Hacker
> Software Developer
> 1631 8th St.
> New Orleans, LA
> 70115
>



-- 
Kurtis Rainbolt-Greene, Hacker
Software Developer
1631 8th St.
New Orleans, LA
70115

krainboltgreene:
So I've decided to not continue this as I'm having a rather frustrating time with Node. I'm going to try my hand at a Ruby version. Thanks for looking into this.

Issue #71, opened 2015-06-16
Building Mac applications on Windows generates no output


Stephn-R:
``` shell
electron-packager . ${APP_NAME} --platform=darwin --arch=x64 --version=${ELECTRON_VERSION} --out=${OUTPUT_DIR}/mac64
```

Apparently the above command doesn't output anything

maxogden:
Hi, can you answer these questions to help us triage your issue? https://github.com/maxogden/electron-packager/blob/master/CONTRIBUTING.md

Stephn-R:
Version: 4.1.3
CLI Arguments:
    - Platform: Darwin
    - Arch: x64
    - Version: 0.28.1
    - Out: /dist
Platform: Building for Mac

There was no stack trace.

jden:
@Stephn-R to clarify, are you building *on* a mac as well?

Stephn-R:
no I am building on a windows machine

normmcgarry:
+1 same issue, building mac on Windows. same params.

When running it with node, I get this error in the callback:
    { [Error: EXDEV, rename 'C:\Users\User\AppData\Local\Temp\electron-packager-mac\
AppName.app']
      errno: 52,
      code: 'EXDEV',
      path: 'C:\\Users\\User\\AppData\\Local\\Temp\\electron-packager-mac\\AppName.app' }`

jden:
Ah, looks like a cross-device error which we've run into before. @normmcgarry is there a stack trace or a line number associated with that?

normmcgarry:
Looks like `electron-packager\mac.js:83:27`

```javascript
Trace: { [Error: EXDEV, rename 'C:\Users\Norm\AppData\Local\Temp\electron-packag
er-mac\Hook Flash MOV Exporter.app']
  errno: 52,
  code: 'EXDEV',
  path: 'C:\\Users\\Norm\\AppData\\Local\\Temp\\electron-packager-mac\\Hook Flas
h MOV Exporter.app' }
    at s:\Repositories\Internal\Development\hook-internal-repos.internal-tools-f
lash-mov-exporter\trunk\app\index.js:19:15
    at moved (s:\Repositories\Internal\Development\hook-internal-repos.internal-
tools-flash-mov-exporter\trunk\app\node_modules\electron-packager\mac.js:83:27)
    at Object.oncomplete (fs.js:107:15)
```

malept:
Looks like #60. I'll file a PR later today.

[...]

PR #88, opened 2015-06-28
Implement --all, refactor code, add tests

### LINKED ISSUE #38: `--out` inconsistency
If I specify `--out=dist` it will put the contents in:

- OS X: `dist/Foo.app/`
- Windows: `dist/Foo-win32/`
- Linux: `dist/`

Notice how on Linux the contents is put directly into the `--out` folder. This looks like a bug.

### LINKED ISSUE #40: --all option
if you pass `--all` it should compile all possible combinations of arch and platform (that make sense)

### LINKED ISSUE #70: Issue with outputting Win32 versu Win64 apps
```shell
electron-packager . ${APP_NAME} --platform=win32 --arch=ia32 --version=${ELECTRON_VERSION} --out=${OUTPUT_DIR}
```

```shell
electron-packager . ${APP_NAME} --platform=win32 --arch=x64 --version=${ELECTRON_VERSION} --out=${OUTPUT_DIR}
```

Both of these calls export the same root folder ```${APP_NAME}-win32```

Shouldn't it be expected that the Win64 version would have 64 as a suffix?

### LINKED ISSUE #77: we need some tests!
lots of people are using this but nobody has written a test suite yet to catch regressions!

guidelines:

- we would prefer https://www.npmjs.com/package/tape
- keep things simple, focused on catching regressions. we can figure out proper x-platform dockerized integration testing etc in the future.
- our current 'test' command just [runs standard](https://github.com/maxogden/electron-packager/blob/554f3b735483319450c2ad8d0e0cba90e041ebe2/package.json#L35). make sure to keep that (e.g. `standard && tape tests/*.js`)

### LINKED ISSUE #84: Should ignore comparisons trim opts.dir first?
All of the instances in which `common.userIgnoreFilter` is called involve `ncp` operations to copy `opts.dir` and exclude certain patterns.  However, the exclusion logic in `userIgnoreFilter` currently matches each pattern against each full path.

This can lead to frustrating bugs depending on the full path under which you are working.  For example, if you have an ignore filter for `'/electron/'`, but your entire structure is under `/Users/myuser/dev/electron/myproject/`, then *nothing* will be copied into your application (due to everything matching the *parent* `electron` directory).  This same problem won't occur if you were to run under, say, `/Users/myuser/dev/js/myproject/` instead.

It seems to me that it'd be generally safer and more useful if `userIgnoreFilter` trimmed `opts.dir` from the front of each filename it compares.  This would also make it feasible to use patterns starting with `^` to match top-level directory entries within the app directory, which in itself can be useful for cases where you want to match a top-level directory but *not* directories of the same name underneath other directories (e.g. a top-level `util/` but not `some/other/lib/util/`).

My only reservation is that this change would be somewhat backwards-incompatible (i.e. for strings such as `/dirname/` expecting to match a "top-level" directory under the app directory right now).

kfranqueiro:
This adds support for --all, --platform=all, and --arch=all.

In order to accommodate outputting multiple directories for multiple platforms and architectures, this also implements a new directory structure under the output folder (distinguished by both platform and
arch).  This structure is applied even to OS X distributions, which formerly were output directly to an .app folder.  This could be considered a backwards-incompatible change.

One other backwards-incompatible change is the value that the packager function passes to the callback, which is now always an array of paths, rather than just a single path.

The behavior of the icon option has also been modified to use its basename and apply .ico or .icns depending on platform, to make it usable with --all and --platform=all.  This attempts to maximize backwards compatibility, by allowing a full filename to be specified, but replacing the filename's extension with what is appropriate for each target platform.  Alternatively, the extension can now be omitted.

In the process of implementing this, it became evident that some things were being done in 3 different places, and weren't always being done consistently, so I've deduplicated everything I could.

This also includes a few other changes to improve stability for multi-target runs, and other fixes:

* Avoid targeting darwin if the environment doesn't support symlinks,
  to avoid the process bailing out on Windows
* Implement --overwrite centrally in index.js such that it explicitly
  skips if an output directory already exists, for consistency with
  all target platforms and to avoid any possible errors that would halt
  operation during one target of a multi-target run
* Use ncp instead of mv to move to finalPath, which avoids flakiness
  I noticed when testing on Windows 8 especially with multi-target runs
* Simplify temp directory logic by using a nested structure, so there
  is only one top-level directory to clean up
* Reinstate fix from #55 which seems to have been clobbered by a later
  merge
* linux.createApp now resolves to the final output directory;
  it was formerly resolving to the executable path
* mac.createApp now replaces space with underscore in bundle IDs
* Only the platform modules that are needed are loaded
* The win32 module only loads rcedit if needed

This also fixes a couple of missing updates to docs (readme/usage).

The added tests cover the following cases:

* Basic runs for a specific version/platform/arch with default options
  (testing for existence of expected output files/folders)
* Runs for each target platform with `out` set
* Runs for each target platform with `asar` set
* Runs for each target platform with `prune` set
* Runs for each target platform with `ignore` set
  * This includes testing `ignore` including a path separator
  * This also includes testing that `ignore` entries are applied
    with the app folder trimmed
* Runs for each target platform with `overwrite` set / not set
* Runs with `all` set, with `arch` or `platform` each set to `all`,
  and with `platform` set to multiple platforms (comma-delimited or
  array)
* Tests specifically for the darwin target platform, to test
  `icon` and `sign`

The test logic is organized such that the version of Electron being tested can be configured once centrally (in config.json), and the tests will not run until after downloading any necessary electron versions (to avoid timing out due to long downloads).

This addresses the following issues:

* Resolves #40
* Resolves #38
* Resolves #70
* Works around #71
* Resolves #84 by reinstating #55
* Resolves #77 with tests


kfranqueiro:
I've run the tests on OS X 10.10, Xubuntu, Windows 7, and Windows 8, and verified they passed before submitting this PR.

maxogden:
EPIC

maxogden:
published as 5.0.0 and added @kfranqueiro as a collaborator. thanks a ton for this!

kfranqueiro:
Awesome!  Thanks for accepting it.  Glad my OCD paid off :)

PR #93, opened 2015-07-02
Adds CFBundleShortVersionString to the main .plist on OSX

johansatge:
Hi,

this PR adds the `CFBundleShortVersionString` key to the `.plist` file when building an OSX app.

This key is used when displaying the app version in the Finder.

Currently, without the key:

vuplicity-darwin-x64 2015-07-02 15-24-05

With the key:

vuplicity-darwin-x64 2015-07-02 15-24-05 copy

Thanks !

### DIFF
--- a/mac.js
+++ b/mac.js
@@ -33,6 +33,7 @@ module.exports = {
 
       if (appVersion) {
         appPlist.CFBundleVersion = appVersion
+        appPlist.CFBundleShortVersionString = appVersion
       }
 
       if (opts.protocols) {


armorik83:
:+1:
We also needed.

uiureo:
+1

malept:
Thanks for the contribution, but I'm more inclined to take #127 instead, as it points out the differences between the two version keys, and also contains doc/test changes.

johansatge:
Thanks for your comment!

Indeed @herrmannplatz's solution in #127 makes more sense.

PR #94, opened 2015-07-02
Inferred name and version from package.json

### LINKED ISSUE #37: Infer from package.json
Would be useful if it could infer the name from the `name` field in package.json (titelized) and the `--version` from the "electron-prebuilt" dependency. This to reduce config and not requiring to define this in multiple places. This is just default and users could just override as normal with the cli arguments.

rahatarmanahmed:
EDIT: updated description of how version is inferred.

The name argument and version option are no longer required. If not specified, it will search upwards from opts.dir for package.json files. It picks up the app name from the packageName or name properties, and the ~~version~~ installed electron-prebuilt location from devDependencies.electron-prebuilt or dependencies.electron-prebuilt. However, they don't necessarily have to come from the same package.json, it will use the first available value from the nearest package.json.

~~Because package.json dependencies have version ranges, it will also query npm for the versions that satisfy that version range. It uses the latest version in the range.~~

~~An indirect change is that versions specified by an explicit option are still tested as a version range, so using something like --version ^0.28.1 is now possible.~~

Once the electron-prebuilt location is found, it looks in `node_modules/electron-prebuilt/package.json` to get the final version to use.

This should not cause any breaking changes, as previously provided versions should still resolve to the same version.

Resolves #37


rahatarmanahmed:
I haven't yet added tests or documentation changes yet because I want to be sure these defaults are okay, first.

rahatarmanahmed:
Someone brought to my attention that some more options could be inferred from package.json, like "os" and "cpu". I'll probably leave that for someone else to implement, as I don't need it right now, and trying to figure out how to resolve the whitelist/blacklist of those properties is non-trivial.

kfranqueiro:
This is neat, but I have a few concerns about the approach of relying on npm + semver:

* This just hangs if you try to run it without an internet connection
* Adding npm as a dependency increases the payload of this repository's dependencies by 17MB
* If you use a version range, you may unwittingly end up packaging with a different version than you've been testing with in development (for example, if you used `~0.28.2`, never updated your dependencies, but then ran packager after 0.28.3 released)

I am wondering if it would be more appropriate, once you find reference of `electron-prebuilt` in a `package.json`, to then inspect `node_modules/electron-prebuilt/package.json` relative to the directory where you found it specified as a dependency, and used the version that is already installed.  This would avoid the caveats above.

rahatarmanahmed:
I initially did it this way because I thought it would be "more correct", but you bring up good points. I'll look into npm dedupe behavior (especially what's coming in npm@3), but if everything checks out I'll change it to grab the version from electron-prebuilt's package.json.

rahatarmanahmed:
I think one way to avoid the npm dedupe issue is to use the same code I used to search upwards for `package.json` to also search for `node_modules/electron-prebuilt/package.json`. The idea being that even if electron-prebuilt is deduped and is installed in a parent directory, the search will be guaranteed to find it?

From my understanding of deduping, that should work. But it's late so I might be missing something obvious.

rahatarmanahmed:
Alright, changes have been made as described above. Works fine for me, tests still pass. Later I'll add tests and documentation after some eyes have checked it.

index.js
```diff
@@ -149,21 +214,26 @@ function createSeries (opts, archs, platforms) {
...
-    if (err) return cb(err)
-
-    cb(null, appPaths.filter(function (appPath) {
-      // Remove falsy entries (e.g. skipped platforms)
-      return appPath
-    }))
+  getNameAndVersion(opts, function (err) {
+    if (err) {
+      return cb(err)
+    }
```
malept:
````javascript
if (err) return cb(err)
```
...for consistency.

rahatarmanahmed: Ah, standard yelled at me for not having braces, I didn't realize you could do this. Fixed.

rahatarmanahmed:
Added tests and documentation, rebased everything into one commit. This PR should be ready now if no one has issues with it.

index.js
```diff
@@ -62,6 +62,56 @@ function validateList (list, supported, name) {
...
+        } catch (err) {
+          err.message = 'Error parsing ' + filename + ': ' + err.message
+          return cb(err)
+        }
+      }
+      var parent = path.join(dir, '..')
+      if (path.resolve(parent) === path.resolve(dir)) return cb(null)
+      searchUpwards(name, parent, fn, cb)
+    })
+  }
```
maxogden: I would prefer to use https://www.npmjs.com/package/resolve instead of this function

maxogden:
@rahatarmanahmed other than the 1 comment I made above this looks good to me. Inferring `name` and `version` seems like a good user friendly idea. My only concern would be making it too 'magic' but this seems OK since if your app doesnt turn out how you want you can just go back and specify the `name` and `version` later.

rahatarmanahmed:
@maxogden ah, didn't think to look for a package for that! Made the change and the code is much simpler now.

jden:
@rahatarmanahmed can you please rebase & squash the PR into one commit? Thanks :)

rahatarmanahmed:
Done.

maxogden:
Sorry I've been swamped lately, if any of the other maintainers wanna review, merge and publish this and the other PRs feel free :) Otherwise I'll get to it eventually

malept:
The only comment I have is that it might be useful to add a test for the directory traversing part of `getNameAndVersion`. Apart from that, :+1:

rahatarmanahmed:
Sorry, I didn't notice that last comment! I'm rebasing this PR with the latest master and making sure tests pass.

As far as testing `getNameAndVersion`, perhaps it'd be best to put that kind of logic in a separate module with tests ? (i've seen the light RE: small module philosophy.)  I'm thinking a `get-package-info` module that will traverse parent package.json's as in this PR, returning requested properties.

rahatarmanahmed:
@maxogden @malept are the tests currently passing on master? Strange to see travis failing on node 0.12 and 5, but not 4.x. Seems like previous documentation only PRs have had failing tests, too?

rahatarmanahmed:
Ah, looks like the issue is that tests are timing out before they can finish. Raising the timeout fixed that on my end but I don't know if there was a reason for the timeout being 15s.

malept:
Thanks for working on this again!

> As far as testing `getNameAndVersion`, perhaps it'd be best to put that kind of logic in a separate module with tests ? (i've seen the light RE: small module philosophy.) I'm thinking a `get-package-info` module that will traverse parent package.json's as in this PR, returning requested properties.

No objections from me.

> I don't know if there was a reason for the timeout being 15s. 

I'm not aware of a particular reason.

rahatarmanahmed:
Tests are passing, hooray!

- Refactored out some of the logic into a separate module [`get-package-info`](https://github.com/rahatarmanahmed/get-package-info).
- Increased timeout so tests don't fail prematurely
- Fixed a small annoyance where stopping tests mid-test will leave files in `test/work` and screw up future tests. Now it runs `rimraf test/work` before starting a test.

malept:
I'm testing your branch with my own app, and the docs you added indicate that omitting `appname` will make electron-packager choose either `productName` or `name` from `package.json`. I have both keys in my `package.json`, and when I run `electron-packager`, the `name` is used. I would have thought that `productName` would be preferred.

rahatarmanahmed:
Is the `productName` from the same project.json as `name`? I have a test in `get-package-info` to make sure it behaves this way:
https://github.com/rahatarmanahmed/get-package-info/blob/master/test/node_modules/we/need/to/go/deeper/package.json
https://github.com/rahatarmanahmed/get-package-info/blob/master/test/test.js#L49-L50

malept:
Never mind, I forgot to save `package.json` :sheep: 

Works great for me!

PR #99, opened 2015-07-07
fix(mac): handle spaces when signing

imlucas:
currently just fails silently :(

[Review on Reviewable](https://reviewable.io/reviews/maxogden/electron-packager/99)


maxogden:
nice catch, thanks. published as 5.0.1

PR #107, opened 2015-07-24
Add support for Linux ARM

### LINKED ISSUE #106: Please add support for Linux ARM

JamesKyburz:
Fixes #106.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,6 +1,7 @@
 var path = require('path')
 var fs = require('fs')
 var os = require('os')
+var semver = require('semver')
 
 var download = require('electron-download')
 var extract = require('extract-zip')
@@ -11,7 +12,8 @@ var common = require('./common')
 
 var supportedArchs = {
   ia32: 1,
-  x64: 1
+  x64: 1,
+  arm: 1
 }
 
 var supportedPlatforms = {
@@ -68,6 +70,7 @@ function createSeries (opts, archs, platforms) {
     platforms.forEach(function (platform) {
       // Electron does not have 32-bit releases for Mac OS X, so skip that combination
       if (platform === 'darwin' && arch === 'ia32') return
+      if (platform !== 'linux' && arch === 'arm') return
       combinations.push({
         platform: platform,
         arch: arch,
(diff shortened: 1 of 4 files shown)


malept:
ARM builds are only available starting with 0.29.0. Do we want to restrict building that OS/arch combination with `--all` if the app requires a lower version?

malept:
The commit message should be more descriptive.

JamesKyburz:
@malept Sorry about the commit message. I have now added a description :)

How about only excluding arm for versions < 0.29.0 when using the --all flag.

That way when building with --all flag prior to version 0.29.0 no arm build will be attempted.
And an explicit build of --platform linux --arch arm will fail with 404 if used in combination with an unsupported version.

malept:
@JamesKyburz that seems like reasonable behavior to me.

JamesKyburz:
@malept I will update the pull request and let you know when it is ready to check.

JamesKyburz:
@jden I had a look at the tests. It seems at present all the tests run using electron version 0.28.3.

Shall I add a new file that tests both < 0.29.0 and latest?

maxogden:
i think the arch name needs to be more specific than `arm`, because there are different non-compatible arm architectures out there e.g. armv6, armv7, armv8 and also armhf vs armel. I'm guessing the electron build is for armv7 but I dont think they specify anywhere...

malept:
Looks like armv7 (hard float): https://github.com/atom/electron/blob/8f06bd6f3e217dd9d636525dcb91cab86459081a/toolchain.gypi#L13

maxogden:
@malept nice find

iojs uses the format `armv7l` and `armv6l` https://iojs.org/dist/v2.4.0/

kfranqueiro:
`opts.version < '0.29.0'` isn't actually going to do what you think it is, is it?  You'd need `semver` for that.

All you should need to do to update the version of Electron used for tests is bump it in config.json.  We will have to update the --all tests' expected conditions in terms of how many builds are expected to exist.

(I didn't entertain the thought of testing multiple versions in one run because that would take a lot of time to run...)

JamesKyburz:
I have now updated the pull request and added tests.

The default version for tests is now the now the latest version (`0.30.1`) of electron, I hope this is ok?
Plus I added 0.29.0, 0.28.3 to `test/multitarget.js` testing when arm is and isn't available.

@maxogden I haven't changed the name of arm to `armv7l` yet...
I agree that the name should match the architecture and arm is too vague.
However even though electron only supports armv7l it's download file suffix is `linux-arm` not as iojs :(
Either way we will have to change something if electron supports more that one arm architecture.
Shall I leave be or change to `armv7l`?

maxogden:
@JamesKyburz im guessing this is oversight on part of Electron, i'll open an issue there

maxogden:
ok opened issue: https://github.com/atom/electron/issues/2343. i'd be curious to see what they say

JamesKyburz:
@maxogden nice!

Ideally if electron adopt the iojs way, it would be nice if they could also add the `armv7l` files retrospectively.

Will wait and see!

maxogden:
OK looks like we got a thumsbup https://github.com/atom/electron/issues/2343 but they havent switched over to the new system yet.

miladsajjadi:
is there any update for this issue?

malept:
We're still waiting on a volunteer to submit a change for electron/electron#2343.

PR #108, opened 2015-07-27
Allow app to be built from already packaged ASAR

Ivshti:
This PR allows apps to be built from pre-packaged ASAR's.
The way it works with the CLI is simple: just pass the ASAR in place of the directory.
For programmatic usage, use "asar-package" arg.

Useful when building for a few platforms at once :)

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -4,7 +4,8 @@ var args = require('minimist')(process.argv.slice(2), {boolean: ['prune', 'asar'
 var packager = require('./')
 var usage = fs.readFileSync(__dirname + '/usage.txt').toString()
 
-args.dir = args._[0]
+if (fs.statSync(args._[0]).isDirectory()) args.dir = args._[0]
+else args['asar-package'] = args._[0]
 args.name = args._[1]
 
 var protocolSchemes = [].concat(args.protocol || [])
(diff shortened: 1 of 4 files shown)


Ivshti:
Can we please bump & publish when this is merged? :)

malept:
Could you rebase your commits as appropriate?

Ivshti:
Done.

kfranqueiro:
This should ideally include updates to the rest of the README to indicate that either `dir` or `asar-package` is now required.  (Currently it lists `dir` as required.)

Ivshti:
Thanks for the tip, updated the documentation

malept:
Is it possible to add a test for this?

Apart from that (and a further rebase), it seems like this can be merged.

maxogden:
I agree a test is needed here so that we can catch regressions in the future

malept:
It's been over 6 months since there's been any activity on this PR and it's very out-of-date. I still don't have any objections to adding this feature, however the following things need to be done before it's ready for review:

* Rebase the commits on master
* Instead of `opts['asar-package']`, there is now a merged `asar` set of options (as of #417) so it should be moved into `opts.asar.package`.
* Add at least one test

I'm closing this PR until someone volunteers to pick it up again.

Issue #112, opened 2015-08-06
Support for specifying cache location


quicksnap:
It could be helpful to be able to override the `cache` option that `electron-download` exposes. Would you accept a PR for piping the option through?

PR #113, opened 2015-08-06
Pipe optional cache option to electron-download

quicksnap:
Possible PR for #112 but might need some doc changes.

### DIFF
--- a/index.js
+++ b/index.js
@@ -71,7 +71,8 @@ function createSeries (opts, archs, platforms) {
       combinations.push({
         platform: platform,
         arch: arch,
-        version: opts.version
+        version: opts.version,
+        cache: opts.cache
       })
     })
   })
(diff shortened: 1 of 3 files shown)


quicksnap:
Looking at code, probably fine to just add the key--if it's `undefined` it won't affect `electron-download`

malept:
Definitely needs some docs.

quicksnap:
I can add those, but LMK if you'd accept something like this first =)

quicksnap:
Also, for context, I desire this because we'll be rolling a customized electron build, and overriding the cache would make things easier for me.

malept:
Seems reasonable to me. We could use it to speed up the Travis tests, too.

quicksnap:
Cool--I'll write up changes tomorrow.

quicksnap:
Updated PR with documentation and adjustment to the implementation

quicksnap:
Any update? Good to merge? =)

Issue #90, opened 2015-07-01
Linux icon doesn't appear


alexwarren:
Are there any special requirements for Linux icons? My understanding is that a PNG file should work. But when I pass a PNG file into the `--icon` option, it doesn't appear in Ubuntu. The icon remains as a default "?".

I've tried various sizes - 1024x1024, 128x128, 96x96, but I see the same behaviour.

I'm using `--arch=x64` with Ubuntu 14.04, packaging Electron v0.27.3.

malept:
Is this Unity?

alexwarren:
@malept Not sure I understand the question, what does Unity have to do with electron-packager? It's not something I'm using.

malept:
I don't understand where you're trying to see the icon. Perhaps you could give a screenshot?

alexwarren:
Oh I see Unity is the name of a graphical shell. Well... I dunno, I'm just using whatever Ubuntu's default is.

alexwarren:
This is the icon that appears in the app launcher on the left when the app is running:

![screen shot 2015-07-01 at 20 58 12](https://cloud.githubusercontent.com/assets/418188/8463634/ea02237e-2033-11e5-96bc-947c0e872523.png)

malept:
Have you set the icon of the `BrowserWindow` in your Electron app, per http://electron.atom.io/docs/v0.28.0/api/browser-window/ ?

alexwarren:
Ah, fantastic, didn't realise that needed to be set there (as it works without it on Windows and OS X). Thanks!

grofit:
I do not get an icon in the filesystem even with an icon set, just to confirm should the format of the icon be an image format, i.e. `icon: "my-icon.png"` or an ico file `icon: "my-file.ico"` as I am using the latter and it works fine in windows, but linux shows nothing.

malept:
I would imagine that using a `.ico` as the file format would not work with Unity.

strongant:
var iconPath = __dirname + '/app/assets/img/appledisk.png';

function createWindow() {
  mainWindow = new BrowserWindow({
    width: 1000,
    height: 1000,
    icon: iconPath
  });

when electron start, i see nothing.[https://cloud.githubusercontent.com/assets/418188/8463634/ea02237e-2033-11e5-96bc-947c0e872523.png](url)

malept:
@strongant check the application bundle and make sure the icon path you specify is relative to the JavaScript file you're calling it from.

[...]

PR #125, opened 2015-09-03
Document Icon Filetypes 

jlord:
:wave: 

You're sitting right next to me and saw me type these things so I hope you like it :pizza: :cat2:

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -105,7 +105,11 @@ Shortcut for `--arch=all --platform=all`
 
 `icon` - *String*
 
-For OS X, this must be `.icns` format.
+  Currently you must look for conversion tools in order to supply an icon in the
+  format required by the platform:
+  
+  - OS X: `.icns`
+  - Windows: `.ico`
 
 `app-bundle-id` - *String*
 
--- a/readme.md
+++ b/readme.md
@@ -22,20 +22,20 @@ npm i electron-packager -g
 
 ```
 Usage: electron-packager <sourcedir> <appname> --platform=<platform> --arch=<arch> --version=<version>
-  
+
 Required options
 
 platform           all, or one or more of: linux, win32, darwin (comma-delimited if multiple)
 arch               all, ia32, x64
 version            see https://github.com/atom/electron/releases
-                  
+
 Example            electron-packager ./ FooBar --platform=darwin --arch=x64 --version=0.25.1
 
 Optional options
 
 all                equivalent to --platform=all --arch=all
 out                the dir to put the app into at the end. defaults to current working dir
-icon               the icon file to use as the icon for the app
+icon               the icon file to use as the icon for the app. Note: Format depends on OS.
 app-bundle-id      bundle identifier to use in the app plist
 app-version        version to set for the app
 cache              directory of cached electron downloads. Defaults to '$HOME/.electron'
@@ -57,7 +57,7 @@ version-string     should contain a hash of the application metadata to be embed
                    - ProductVersion
                    - ProductName
                    - InternalName
-                   
+
 
 ```
 
@@ -104,6 +104,8 @@ Shortcut for `--arch=all --platform=all`
 `out` - *String*
 
 `icon` - *String*
+
+For OS X, this must be `.icns` format.
 
 `app-bundle-id` - *String*
 


RIAEvangelist:
Do we have to do a special build with  the .icns format just for OSX or can we still do a build for all platforms with that format? If a special build is needed it would make sense to add a conversion tool, like png to icns to support uniform implementation when the 'all' member is set to true, or non OSX builds are mixed with OSX builds.

readme.md
```diff
@@ -105,6 +105,8 @@ Shortcut for `--arch=all --platform=all`
 
 `icon` - *String*
 
+For OS X, this must be `.icns` format.
+
```
malept: Might as well mention what the file format is for all platforms.

malept:
@RIAEvangelist it's a little bit more complicated than that. [In my app](https://github.com/malept/gmusicprocurator-app/blob/938498bf4be0c3d5c8960d501b91e47ed31dc0b9/Makefile#L61-L66), I generate `.icns` and `.ico` files from a `.svg` file, and found that the PNG :arrow_right: ICNS tool depends on what the build OS is.

readme.md
malept: (And that you don't need to set this if you're just building for Linux, see #90.)

maxogden:
@RIAEvangelist I would guess that a module that did icns, ico, png and svg conversion would be pretty heavy. I think the simplest thing we could do is have e.g. `--mac-icon`, `--linux-icon`, `--win-icon` flags (or maybe using the platform name so `--darwin-icon` etc)

kfranqueiro:
Currently the `icon` option automatically strips the extension of the filename and looks for `icns` for Mac and `ico` for Windows.

malept:
@kfranqueiro ...I did not know that. That should be documented :smile:

readme.md
RIAEvangelist: agree @ mention file format.

Oh yeah, I forgot #90 is how I got my linux apps to work. Does this work on all platforms as well? I know it wont update the binary logo on those systems, but it would be smart to always do that if so.

RIAEvangelist:
Wow @kfranqueiro ! Sweet knowledge share there! That is indeed very important to be documented.

kfranqueiro:
It was implemented as part of #88, since otherwise `--all` with `--icon` didn't make a whole lot of sense.  Looking at how it's doc'd (or not), yeah, I should've thought to expand on it.  I was probably not sure of how much I should touch docs at the time.

jlord:
I updated the PR to have a list stating the OS X and Windows filetypes. 

I have no idea how Linux is supposed to be done? Do users need a `.desktop` file?

malept:
Linux just needs a `.png`, and it doesn't need to be set in `--icon`, per my comments in https://github.com/maxogden/electron-packager/issues/90#issuecomment-117808176.

jlord:
@malept Hmm. I read that thread but it wasn't clear to me and I had tried with a PNG. But I will try again not passing in `--icon` at all and just relying on the png in `BrowserWindow` .

malept:
@jlord I think I need some clarification on the behavior you're expecting. AFAIK, in Linux file managers, you can't embed an icon in the executable and have it show up when it's in the executable list. The `BrowserWindow` setting is required to show the icon in the window list (unlike, IIRC, OS X).

jlord:
@malept the thing I'm testing/trying for is an app icon in the file manager and dock on Ubuntu.

RIAEvangelist:
Heres a snippit from my video player :

RIAEvangelist:
Here's a snippit from my [video player](https://github.com/RIAEvangelist/electron-video-player) its in [main.js](https://github.com/RIAEvangelist/electron-video-player/blob/master/src/main.js) :



    app.on(
        'ready',
        function() {
            player = new BrowserWindow(
                {
                    title:'Electron Video Player',
                    'accept-first-mouse':true,
                    width: 640,
                    height: 480,
                    'min-width': 640,
                    'min-height': 480,
                    frame:false,
                    icon:__dirname+'/img/logo-256.png',
                    'text-areas-are-resizable':false
                }
            );

            player.loadUrl('file://' + __dirname + '/player.html');

            //player.openDevTools();

            player.on(
                'closed',
                function() {
                    player = null;
                }
            );
        }
    );

you can see where I set the linux icon.

jlord:
@RIAEvangelist Hmm, really not sure what I'm doing differently but that doesn't work for me. I'll open another issue for this specifically so we can clear it up and add it to the docs.

Otherwise @maxogden this PR is good to go for at least specifying the filetypes for OS X and Windows.

PR #127, opened 2015-09-04
Add build version number for OSX

herrmannplatz:
Like in #93 i need to add the release and build version number for my OSX app. Unfortunately that PR wouldn't solve my problem cause i need to add a different build version.

Also the `app-version` is stored in the `CFBundleVersion ` property which is, according to [Apple](https://developer.apple.com/library/ios/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html), meant to store the build version number, whereas the `CFBundleShortVersionString ` stores the actual release version. I would suggest to store the `app-version` in `CFBundleShortVersionString` and the new `build-version` in `CFBundleVersion`.

Besides that great project. Whole setup worked like a charm.

### DIFF
--- a/mac.js
+++ b/mac.js
@@ -43,6 +43,7 @@ module.exports = {
       // Update plist files
       var defaultBundleName = 'com.electron.' + opts.name.toLowerCase().replace(/ /g, '_')
       var appVersion = opts['app-version']
+      var buildVersion = opts['build-version']
 
       appPlist.CFBundleDisplayName = opts.name
       appPlist.CFBundleIdentifier = opts['app-bundle-id'] || defaultBundleName
@@ -51,7 +52,11 @@ module.exports = {
       helperPlist.CFBundleName = opts.name
 
       if (appVersion) {
-        appPlist.CFBundleVersion = appVersion
+        appPlist.CFBundleShortVersionString = appPlist.CFBundleVersion = appVersion
+      }
+
+      if (buildVersion) {
+        appPlist.CFBundleVersion = buildVersion
       }
 
       if (opts.protocols) {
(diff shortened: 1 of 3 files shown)


lipis:
+1

malept:
Speaking as someone who doesn't know much about the OS X developer ecosystem, is it appropriate if `CFBundleVersion` defaults to `app-version`, if `build-version` is not specified?

herrmannplatz:
I don't know if its appropriate.

I use hockey and their [recommended way](http://support.hockeyapp.net/kb/client-integration-ios-mac-os-x/how-to-do-versioning-for-beta-versions-on-ios-or-mac-os-x) for creating versions, in order to distribute beta builds. In this case release version and build version are completely different.

lipis:
@malept any chance we can merge that rather sooner than later?!

malept:
@lipis I would be more comfortable merging this if a contributor who is more familiar with building apps for OSX than I am were to comment on it. It looks fine, purely from a code perspective.

lipis:
@jaanus Maybe you can help us?

maxogden:
@lipis you can easily install herrmannplatz's fork using npm and use that immediately

lipis:
@maxogden we might do that eventually if that PR is going to take some time (because it will get messy with more people working on the same project)..

jaanus:
Hey guys, I’ve been working a lot on iOS/OSX apps and also have dealt a lot with the version numbering questions. Yes, what @herrmannplatz says is appropriate about them — `CFBundleVersion` should be an increasing number, and `CFBundleShortVersionString` should be the “actual release version”, which also matches how Hockey describes their usage. And…

> is it appropriate if CFBundleVersion defaults to app-version, if build-version is not specified?

Yes, it’s fine. These two can be set to the same string if you don’t wish/bother to specify a separate `build-version`.

herrmannplatz:
I updated my PR. It now set `app-version` and `build version` to the same value if no `build-version` specified

malept:
Thanks, @jaanus.

@herrmannplatz, once you rebase and the tests pass, I'll merge.

herrmannplatz:
Should be fine now

PR #130, opened 2015-09-11
upgrade asar

fscherwi:


maxogden:
@fscherwi could you say more about why asar needs upgrading, e.g. does it fix an open issue?

malept:
A similar change was introduced in #174.

PR #132, opened 2015-09-16
Fixate dependency versions

soygul:
Otherwise sub dependencies break frequently, which breaks builds on platforms randomly (which is the case with linux/windows builds with latest version as of today, but was working 2 days ago, and only thing changed were minor sub dependency versions).

This PR fixes deps to their original versions, which I've verified to work on all platforms in our build system.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,17 +17,17 @@
   },
   "homepage": "https://github.com/maxogden/electron-packager",
   "dependencies": {
-    "asar": "^0.6.1",
-    "electron-download": "^1.0.0",
-    "extract-zip": "^1.0.3",
-    "minimist": "^1.1.1",
-    "mkdirp": "^0.5.0",
-    "mv": "^2.0.3",
-    "ncp": "^2.0.0",
-    "plist": "^1.1.0",
-    "rcedit": "^0.3.0",
-    "rimraf": "^2.3.2",
-    "run-series": "^1.1.1"
+    "asar": "0.6.1",
+    "electron-download": "1.0.0",
+    "extract-zip": "1.0.3",
+    "minimist": "1.1.1",
+    "mkdirp": "0.5.0",
+    "mv": "2.0.3",
+    "ncp": "2.0.0",
+    "plist": "1.1.0",
+    "rcedit": "0.3.0",
+    "rimraf": "2.3.2",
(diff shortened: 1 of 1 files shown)


stefanjudis:
The correct way for this would be to use `npm shrinkwrap`, or?

soygul:
`shrinkwrap` fixes to latest possible minor versions (which is the problem here as they are currently broken)

stefanjudis:
> shrinkwrap fixes to latest possible minor versions (which is the problem here as they are currently broken)

Nope, it fixes the installed versions recursively down the whole dependency tree. Means a shrinkwrap file of a working version fixes the whole tree.

Setting only direct dependencies to a specific version does not help, when the dependencies of the dependencies go with `^` or `~`.

soygul:
I was talking about a clean build agent for instance: `git clone ... npm install ... npm shrinkwrap` and you have latest broken deps of `^` line. Fixing the first layer of dependencies at least provides basic defense.

soygul:
bundling the updating entire shrinkwrap file inside this repo would be a big burden for the developer..

stefanjudis:
> bundling the updating entire shrinkwrap file inside this repo would be a big burden for the developer..

For safety - I'd say this is better than going with a half-defence. :)

@maxogden for sure has a strong opinion on that. 

And sorry for jumping it - this topic is really interesting for me. ;)

malept:
Could you perhaps try just locking the extract-zip dependency locally and see if that fixes the problem? @maxogden updated that yesterday.

soygul:
Might be the problem: https://github.com/maxogden/extract-zip/issues/10

Didn't try myself though, already using my own fork atm with the fixed deps.

maxogden:
in this case the breakage was my fault, I only tested on Mac OS. in the future i'll be sure to test on windows and linux before upgrading dependencies. most of the time having `^` saves a lot of time and frustration so I'm keeping them, and will put in place better testing infrastructure to prevent issues here in the future.

soygul:
:+1:

maxogden:
By the way, if anyone here wants to help with the `extract-zip` module by improving its test suite I opened an issue for adding the electron zips to the `extract-zip` test suite here: https://github.com/maxogden/extract-zip/issues/11

PR #143, opened 2015-09-27
Fix #129: support file-version metadata (win32)

### LINKED ISSUE #129: version-string.FileVersion is not working on Windows
When I use this command

`electron-packager c:\myapp MyApp --platform=win32 --arch=x64 --version=0.32.1 --app-version=1.0.0.123 --version-string.ProductVersion=1.0.0.123 --version-string.FileVersion=1.0.0.123`

then the *File version* of MyApp.exe contains 0.32.1 instead of 1.0.0.123, see the attached screenshot.

*Windows 7 x64
electron-packager v5.1.0*

electron-packager-fileversion

spinico:
I had trouble cleaning up my commit logs on previous pull request. I'm not sure how to put up a test for windows executable file version though.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,8 @@
   "devDependencies": {
     "run-waterfall": "^1.1.1",
     "standard": "^3.3.2",
-    "tape": "^4.0.0"
+    "tape": "^4.0.0",
+    "rcinfo": "^0.1.3"
   },
   "scripts": {
     "test": "standard && tape test"
--- a/test/index.js
+++ b/test/index.js
@@ -19,6 +19,7 @@ series([
   console.log('Running tests...')
   require('./basic')
   require('./multitarget')
+  require('./win32')
 
   if (process.platform !== 'win32') {
     // Perform additional tests specific to building for OS X
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,7 @@ node_js:
 cache:
   directories:
     - $HOME/.electron
+addons:
+  apt:
+    packages:
+    - wine
--- a/test/win32.js
+++ b/test/win32.js
@@ -1,0 +1,54 @@
+var fs = require('fs')
+var path = require('path')
+
+var packager = require('..')
+var test = require('tape')
+var waterfall = require('run-waterfall')
+
+var config = require('./config.json')
+var util = require('./util')
+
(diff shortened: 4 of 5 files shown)


malept:
I would add a `test/win.js`, just like there's a `test/mac.js`.

spinico:
Added test file are required only on win32 platform. It uses native win32 functionalities to retrieve exe file metadata. For this, I needed to add a new dev dependency (rcinfo) to package.json.

package.json
```diff
@@ -25,14 +25,15 @@
     "mv": "^2.0.3",
     "ncp": "^2.0.0",
     "plist": "^1.1.0",
-    "rcedit": "^0.3.0",
+    "rcedit": "^0.3.0",
```
malept: Please don't add trailing whitespace.

test/index.js
```diff
@@ -20,8 +20,10 @@ series([
   require('./basic')
   require('./multitarget')
 
-  if (process.platform !== 'win32') {
-    // Perform additional tests specific to building for OS X
+  // Perform additional tests specific to OS
+  if (process.platform === 'darwin') {
```
malept: `process.platform === 'darwin'` is not the same as `process.platform !== 'win32'`.

test/index.js
```diff
@@ -20,8 +20,10 @@ series([
...
     require('./mac')
+  } else if (process.platform === 'win32') {
+    require('./win32')
```
malept: Is it possible to run these tests via wine (particularly on Linux, so we can integrate them into the repository's continuous integration)?

test/index.js
spinico: Ok, I'll revert that change.

test/index.js
spinico: I'll have a look, should be possible.

package.json
spinico: Sorry about that.

spinico:
For now the CI build fails when attempting to use wine to run the win32 tests. Is there something to setup on travis for this to work?

malept:
Yeah, wine needs to be installed first: https://github.com/malept/gmusicprocurator-app/blob/master/.travis.yml#L14

webarchymeta:
I works for me

Issue #68, opened 2015-06-15
Ignore the output directory when possible


malept:
Examples of the problem: #64, #67

develar:
Obsolete since #155 ?

malept:
Yep, thanks for reminding me.

PR #155, opened 2015-10-13
Automatically ignoring --out path

### LINKED ISSUE #120: `--out` should be added to the ignore list by default
If `--out` is specified and not `cwd` it should be added to the ignore list.

It makes no sense to include the `out` directory, so this would be a sane default and prevent people from mistakenly include it in their built app.

pyro2927:
Fixes #120.

If the output directory is nested within the directory that gets packed (ie: starts with `./`), any files within it area ignored during the `electron-packager` process.  Also moved the `var ignore...` out of the function so it's not run every time the filter is called for a small performance gain.

### DIFF
--- a/common.js
+++ b/common.js
@@ -32,7 +32,12 @@ function generateFinalPath (opts) {
 function userIgnoreFilter (opts) {
   var ignore = opts.ignore || []
   if (!Array.isArray(ignore)) ignore = [ignore]
-  if (opts.out && opts.out.indexOf('./') === 0) ignore.push('^/' + opts.out.substr(2))
+  if (opts.out) {
+    var out = path.join(process.cwd(), opts.out)
+    if (out.indexOf(process.cwd()) === 0) {
+      ignore.push('^' + out.substr(process.cwd().length))
+    }
+  }
   return function filter (file) {
     file = file.split(path.resolve(opts.dir))[1]
 
--- a/common.js
+++ b/common.js
@@ -30,6 +30,9 @@ function generateFinalPath (opts) {
 }
 
 function userIgnoreFilter (opts) {
+  var ignore = opts.ignore || []
+  if (!Array.isArray(ignore)) ignore = [ignore]
+  if (opts.out && opts.out.indexOf('./') === 0) ignore.push('^/' + opts.out.substr(2))
   return function filter (file) {
     file = file.split(path.resolve(opts.dir))[1]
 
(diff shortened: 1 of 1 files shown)


malept:
Thanks for taking a crack at this. Could you add a test?

kfranqueiro:
This approach seems a bit naive, since it's completely possible for someone to reference a relative directory without `./` in front at all.  Would using `path.resolve` on the output directory and the cwd and comparing the two be a better idea?

I also suspect it may be possible to normalize this into `ignore` up-front rather than every single time the filter runs, which could be a lot.

pyro2927:
> I also suspect it may be possible to normalize this into `ignore` up-front rather than every single time the filter runs, which could be a lot.

My JS-fu is definitely not as good as my ruby-fu, but isn't that what is happening now?  Previously `var ignore = ...` was being run each time the filter function was called, and now that's only being hit once (as well as the adding in of the ignore dir).  Do you want it normalized in a different spot?

> Would using `path.resolve` on the output directory and the cwd and comparing the two be a better idea?

It would be.  I may be able to redo it with that method.

> Could you add a test?

I can try.  I've been writing JS tests for about two days, so I can't promise greatness.

maxogden:
@pyro2927 thanks for getting this rolling!

My two cents is that the `var ignore` optimization isn't relevant, stuff like that only matters for hot code paths, of which there are none in this module (e.g. there are no things getting called millions of times and/or sharing an event loop with a highly concurrent network server). Probably best to just ignore stuff like that and focus on the functionality of the PR in the future.

kfranqueiro:
@pyro2927 Yes, I think so.  However, it occurs to me that we might need to do some substitution for cwd in Windows (since the filename being compared has all backslashes replaced with forward slashes immediately below, but cwd will similarly have backslashes on Windows).

Meanwhile, point taken with how you moved up `ignore` being set, I missed that before (and forgot `userIgnoreFilter` returns a function, and isn't actually the filter function itself).  To @maxogden's point, I suppose we'd have to time a run with a significant number of files under the source directory to see whether the difference is even noticeable, but if it's noticeable *at all* it seems worth avoiding the performance regression, since I'd only expect applications to get bigger...

malept:
Here is where I believe the status of this PR is:

* [ ] Add tests - one of them should test the current working directory case
* [ ] Squash commits

This also happens to fix #68.

develar:
Hi @pyro2927 Do you have time to add tests or I can do it (since it is required for #251)?

pyro2927:
Sadly I don't have time for tests. My current job is keeping me incredibly busy.

PR #156, opened 2015-10-14
Add `build-version` to opts documentation

PaulLeCam:
Hello,

I got a strange bug using Electron's auto-update that turned out to be because I set the `build-version` option as an number while it should be a string (https://github.com/atom/electron/issues/3067).

This PR simply adds this information to the readme.


malept:
Thanks!

PR #157, opened 2015-10-14
Documentation cleanup

malept:
* Reworded a bunch of things
* Removed the duplicate CLI usage section, added a link to usage.txt
* More consistent formatting

I'll rebase before this gets merged.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1,6 +1,6 @@
 # electron-packager
 
-Package your electron app in OS executables (.app, .exe, etc) via JS or CLI. Supports building Windows, Linux or Mac executables.
+Package your [Electron](http://electron.atom.io) app into OS-specific bundles (`.app`, `.exe`, etc.) via JavaScript or the command line. Supports building Windows, Linux or Mac executables.
 
 *formerly known as atom-shell-packager*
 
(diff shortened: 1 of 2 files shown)


readme.md
```diff
@@ -80,89 +47,102 @@ packager(opts, function done (err, appPath) { })
...
-`icon` - *String*
+  Not required if `all` is used.
 
-  Currently you must look for conversion tools in order to supply an icon in the
-  format required by the platform:
+`version` - *String*
 
-  - OS X: `.icns`
-  - Windows: `.ico`
+  Electron version (without the 'v'). See [Electron releases](https://github.com/atom/electron/releases) for valid versions.
```
maxogden: maybe we should have an example here just to hammer it home to people

readme.md
```diff
@@ -8,69 +8,36 @@ Package your electron app in OS executables (.app, .exe, etc) via JS or CLI. Sup
...
-                   - ProductVersion
-                   - ProductName
-                   - InternalName
+### From the Command Line
 
+Running electron-packager from the command line has this basic form:
 
 ```
+electron-packager <sourcedir> <appname> --platform=<platform> --arch=<arch> --version=<Electron version> [optional flags...]
+```
```
maxogden: maybe have a real world example in addition to the template form, so its somewhat copy-pasteable (will also hopefully illustrate to people correct usage with `--version`)

readme.md
malept: :+1:

readme.md
malept: :+1: I think what I want to do is a sub-section with an example command, which would also explain exactly what would happen. Output directory name, abbreviated tree, etc. Make it painfully obvious.

malept:
Added detailed CLI example and an example Electron version number.

malept:
I'll rebase and merge this tomorrow unless there are any objections.

PR #170, opened 2015-10-24
Add that app-version is a mac-only option

markmiro:
From the source it looks like this is a mac-only feature


malept:
IMO this should be a cross-platform option - in fact, there is both a FileVersion and ProductVersion attribute in Windows. We should replace the `version-string` sub-options with the appropriate cross-platform top-level options.

PR #173, opened 2015-10-26
[win32] Ensure product version is set on the executable

### LINKED ISSUE #172: Windows (10): 'Product version'' remains empty
See #166: that case started with three bullet items, but has since then focused on only the second bullet item (App name not set correctly in Windows EXE metadata).

This case is for the first bullet item: 
- If I look at the file details of the created .exe: -- the 'File version' is still set to the used Electron version, although i specify another version -- the 'Product version' is empty

While there is a related case with a PL for the 'File version' (#129/#143), I tried applying the fixes from #143 manually, but that didn't solve the problem for me.

And since also the 'Product version' is not applied, I suspect something else going on, but no clue what

malept:
This is basically a copy/paste solution from #143. Includes a test.

Attempts to fix #172.

### DIFF
--- a/test/win32.js
+++ b/test/win32.js
@@ -49,6 +49,41 @@ function setFileVersionTest (fileVersion) {
   }
 }
 
+function setProductVersionTest (productVersion) {
+  return function (t) {
+    t.timeoutAfter(config.timeout)
+
+    var appExePath
+    var opts = Object.create(baseOpts)
+    opts['version-string'] = {
+      ProductVersion: productVersion
+    }
+
+    waterfall([
+      function (cb) {
+        packager(opts, cb)
+      }, function (paths, cb) {
+        appExePath = path.join(paths[0], opts.name + '.exe')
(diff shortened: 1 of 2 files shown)


malept:
@p-bakker could you test this fix to see if it fixes your specific case?

(It's still a good change to merge, regardless.)

p-bakker:
Tested it: at first seemed to have no effect at all. Debugged a bit and I figured out I wasn't updating the right electron-packager copy (duh...)

So, after I corrected that mistake, the FileVersion started working, but not the ProductVersion.

Debugged that a bit more and found the issue: rcedit doesn't seem to support setting the ProductVersion to value 1. If I try setting it to 0.9 or 1.1 or any other value, the ProductVersion is properly set, if I use 1 (or 1.0) the ProductVersion is not set.

So, conclusion: the fixes for FileVersion/ProductVersion work, but there is another issue in rcedit

Issue #114, opened 2015-08-09
Unable to use Pepper Flash plugin in packaged app (MacOS X)


wolfflow:
I'm unable to harness PPAPI Flash in app packaged by electron packager under  MAC OS X
OS: MacOS X 10.10 Yosemite
Electron: 0.30.3

Sample app: https://github.com/hokein/electron-sample-apps/tree/master/pepper-flash-plugin

NOTE: Windows and Linux packaged apps work well

subblue:
I was having issues with Flash not working until I added ```'web-preferences': {'plugins': true}``` to the ```new BrowserWindow()`` options. This was with version 0.31.0

wolfflow:
Thanks, but I have this flag set already.
https://github.com/hokein/electron-sample-apps/blob/master/pepper-flash-plugin/main.js#L35

wolfflow:
atom/electron/#2457

asual:
I'm hitting this only when using the "--asar" option.

FrancisVarga:
Hi i have the same issue and i actually remove the **--asar** flag already still flash plugin can't be detected

jespermjonsson:
I had the same issue, but after upgrading electron packager from 4.2.0 to 5.1.0 the "X Helper EH.app" was named with the correct application name and "navigator.plugins" is now populated (even with --asar flag).

FrancisVarga:
@jespermjonsson i'm using:
```
"electron-packager": "^5.1.0",
```

hokein:
Verified it works on electron-package v5.1.0 with  https://github.com/hokein/electron-sample-apps/tree/master/pepper-flash-plugin sample:

```
electron-packager ./ flash --asar --platform=darwin --arch=x64 --version=0.34.0
```

But it won't work with `--asar` option.

patricksebastien:
@hokein My app is working when using npm start, but not when packaging the app. Do I have to copy PepperFlashPlayer.plugin after packaging my app? I tried many folders in Contents... Did you want to say that it is working only without the --asar option?

positlabs:
@patricksebastien I've found that binaries can't be run from inside of an asar package. Try moving the plugin to asar-unpacked dir.

patricksebastien:
@positlabs Yes did that but how to call the plugin, I tried:
`app.commandLine.appendSwitch('ppapi-flash-path', __dirname + '/../app.asar.unpacked/PepperFlashPlayer.plugin');`
not working...

[...]

PR #174, opened 2015-10-27
Support unpack directory

### LINKED ISSUE #171: Add option to match directories for asar unpack
I need an entire folder to not be packed into the archive. Right now, the regex is matched only on the filenames. I tried doing a REGEX for file1|file2 but that did not work. It would be great to be able to ignore an entire directory from packing. Perhaps `--asar-unpack-dir` ?

hokein:
The lastest version(v0.8.2) of asar supports unpack directory, atom/asar#44. It's resonable to implement on electron-packager. 

Fixes #171, and part of #114(We can unpack the `PepperFlashPlayer.plugin` directory on OS X).

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -136,6 +136,9 @@ packager(opts, function done (err, appPath) { })
 
 `asar-unpack-dir` - *String*
 
+  Unpacks the dir to app.asar.unpacked directory whose names exactly match this string. The `asar-unpack-dir` is relative to `dir`.
+  For example, `asar-unpack-dir=sub_dir` will unpack the directory `/<dir>/sub_dir`.
+
 `build-version` - *String*
 
 `cache` - *String*
--- a/readme.md
+++ b/readme.md
@@ -134,6 +134,8 @@ packager(opts, function done (err, appPath) { })
 
 `asar-unpack` - *String*
 
+`asar-unpack-dir` - *String*
+
 `build-version` - *String*
 
 `cache` - *String*
--- a/usage.txt
+++ b/usage.txt
@@ -15,6 +15,7 @@ app-bundle-id      bundle identifier to use in the app plist (darwin platform on
 app-version        release version to set for the app
 asar               packages the source code within your app into an archive
 asar-unpack        unpacks the files to app.asar.unpacked directory whose filenames regex .match this string
+asar-unpack-dir    unpacks the dirs to app.asar.unpacked directory whose names match this string
 build-version      build version to set for the app (darwin platform only)
 cache              directory of cached Electron downloads. Defaults to '$HOME/.electron'
 helper-bundle-id   bundle identifier to use in the app helper plist (darwin platform only)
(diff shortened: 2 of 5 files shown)


malept:
@sleman could you test this PR to see if it does what you're asking for in #171?

malept:
@hokein thanks for the PR. Unfortunately, I just merged another PR which causes a merge conflict with this one. Could you please rebase, after adding an entry about `asar-unpack-dir` to the APi section of the README?

hokein:
@malept  sure, done.

usage.txt
```diff
@@ -15,6 +15,7 @@ app-bundle-id      bundle identifier to use in the app plist (darwin platform on
 app-version        release version to set for the app
 asar               packages the source code within your app into an archive
 asar-unpack        unpacks the files to app.asar.unpacked directory whose filenames regex .match this string
+asar-unpack-dir    unpacks the dirs to app.asar.unpacked directory whose names match this string
```
malept: Does this option also work with a regex? If so, the wording should be similar to `asar-unpack`, that is, `unpacks the dirs to app.asar.unpacked directory whose names regex .match this string`

hokein: No, this option won't work with a regex.

readme.md
```diff
@@ -134,6 +134,8 @@ packager(opts, function done (err, appPath) { })
 
 `asar-unpack` - *String*
 
+`asar-unpack-dir` - *String*
+
```
malept: @hokein per #175, could you give a description here as well?

usage.txt
hokein: Not totally. The value of `asar-unpack-dir` not need to match the beginning of the absolute file path of `sourcedir`. It assumes it is under the `sourcedir`.
For example:
```
electron-packager /your/path/to/app app_name --asar-unpack-dir=dir1
```
This will unpack the `/your/path/to/app/dir1` directory.

readme.md
hokein: Done.

usage.txt
malept: Er, I meant the beginning of the relative path, not the absolute path.

hokein:
> Could you rename dir1 to something like subdir_name, to make it less similar to dir?

Done noe.

usage.txt
hokein: Oh, yes.

readme.md
```diff
@@ -134,6 +134,11 @@ packager(opts, function done (err, appPath) { })
...
+  Unpacks the dir to app.asar.unpacked directory whose names exactly match this string. The `asar-unpack-dir` is relative to `dir`.
+  For example, `asar-unpack-dir=sub_dir` is to unpack `/<dir>/sub_dir` dir.
```
malept: This would read better as:

> For example, `asar-unpack-dir=sub_dir` will unpack the directory `/<dir>/sub_dir`.

hokein: Fixed :+1:

malept:
Thanks for your contribution, and your patience.

PR #182, opened 2015-11-05
bugfix

daxtero:
fixed missing directories when create a package of electron app on OSX

### DIFF
--- a/common.js
+++ b/common.js
@@ -73,7 +73,7 @@ module.exports = {
         })
       },
       function (cb) {
-        mkdirp(tempPath, cb)
+        mkdirp(appPath, cb)
       },
       function (cb) {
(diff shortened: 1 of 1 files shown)


kfranqueiro:
Can you explain a case where this failure occurs?  `appPath` shouldn't need to be created - it should already exist once the template is copied.  (It's a concatenation of `tempPath` and `appRelativePath`, the latter being the path that `app` is found within the electron distribution for the respective OS.)

Have there been any changes to the organization of the electron distributions lately?

malept:
Since there hasn't been a response to the question(s) asked for a few months, I'm going to close this PR. That doesn't mean that we won't accept the change, but we need more information before we can make a decision, and this will likely need a test so that we can detect any regressions of this issue.

PR #206, opened 2015-11-27
Don't include default_app

### LINKED ISSUE #49: default_app and atom.asar remain in Contents/Resources
Would be nice to be able to remove all the default Electron files when building. I'm sure this exists on Windows too, but I haven't looked into what files remain.

inukshuk:
It should not be necessary to include Electron's default_app. (I've only tested this on Linux so far)

Fixes #49

Thanks for electron-packager!

### DIFF
--- a/common.js
+++ b/common.js
@@ -80,6 +80,9 @@ module.exports = {
       },
       function (cb) {
         ncp(opts.dir, appPath, {filter: userIgnoreFilter(opts), dereference: true}, cb)
+      },
+      function (cb) {
+        rimraf(path.join(tempPath, 'resources', 'default_app'), cb);
       }
     ]
 
--- a/common.js
+++ b/common.js
@@ -82,7 +82,7 @@ module.exports = {
         ncp(opts.dir, appPath, {filter: userIgnoreFilter(opts), dereference: true}, cb)
       },
       function (cb) {
-        rimraf(path.join(tempPath, 'resources', 'default_app'), cb);
+        rimraf(path.join(tempPath, 'resources', 'default_app'), cb)
       }
     ]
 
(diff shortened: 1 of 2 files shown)


malept:
I need to set aside some time to test this with my app (since I'm the one who's asserted that a packaged app didn't run on Linux unless you use the `default_app` folder), but could you add a quick test to assert that the default_app folder indeed does not get included?

inukshuk:
Yes, I should have added a test in the first place, sorry. Is it OK to include it with the default tests? Seeing as this is the default behaviour that's where I put it.

Also made this work for the Mac version (where the resource path is different).

For what it's worth, I use this on Linux with no problems whatsoever (not packaging `default_app` that is).

Let me know if I can do anything else!

inukshuk:
Out of curiosity I followed up on this and it turns out that [Electron actually looks for](https://github.com/atom/electron/blob/3a5160f79968998ef214eb1d81c00993b0737637/atom/browser/lib/init.coffee#L72) `app`, `app.asar` and `default_app` in that order.

When you run `electron` with another directory, what actually happens is that `main.js` in default app is executed which then loads your app... in other words, if you experiment with deleting the `default_app` folder and then try to run outside of the resources directory, it won't work (this is what may have happened during your testing). Having said that, since we're packaging the app and therefore know that either `app` or `app.asar` is present it is definitely fine to remove the default app.

Anyway, thought I'd post this in case any of it saves some time.

malept:
> Is it OK to include it with the default tests?

Yep, that's fine.

Also, thanks for your research on Electron load order!

malept:
Finally found some time to check this with my app, it works fine.

PR #209, opened 2015-12-02
Add strict-ssl option

lukeapage:
passes it onto electron-download

### DIFF
--- a/index.js
+++ b/index.js
@@ -72,7 +72,8 @@ function createSeries (opts, archs, platforms) {
         platform: platform,
         arch: arch,
         version: opts.version,
-        cache: opts.cache
+        cache: opts.cache,
+        strictSSL: opts['strict-ssl']
       })
     })
   })
--- a/readme.md
+++ b/readme.md
@@ -169,6 +169,10 @@ If the file extension is omitted, it is auto-completed to the correct extension 
 
 `sign` - *String*
 
+`strict-ssl` - *Boolean*
+
+  Whether SSL certificates are required to be valid when downloading Electron. Defaults to true.
+
 `version-string` - *Object*
 
   Object hash of application metadata to embed into the executable (Windows only):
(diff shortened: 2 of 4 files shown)


index.js
```diff
@@ -72,7 +72,8 @@ function createSeries (opts, archs, platforms) {
         platform: platform,
         arch: arch,
         version: opts.version,
-        cache: opts.cache
+        cache: opts.cache,
+	strictSSL: opts['strict-ssl']
```
malept: The parameter alignment is off by one space.

readme.md
```diff
@@ -169,6 +169,8 @@ If the file extension is omitted, it is auto-completed to the correct extension
 
 `sign` - *String*
 
+`strict-ssl` - *Boolean*
+
```
malept: Could you please add some documentation about this new parameter here?

usage.txt
```diff
@@ -26,6 +26,7 @@ out                the dir to put the app into at the end. defaults to current w
 overwrite          if output directory for a platform already exists, replaces it rather than skipping it
 prune              runs `npm prune --production` on the app
 sign               should contain the identity to be used when running `codesign` (only for building for the darwin platform, on OS X)
+strict-ssl         whether electron should be downloaded with strict-ssl (defaults to true)
```
malept: Capitalize "Electron", and this should probably be reworded to not say `strict-ssl` in the description.

lukeapage:
I ammended the commit with the changes you have suggested.

usage.txt
```diff
@@ -26,6 +26,7 @@ out                the dir to put the app into at the end. defaults to current w
 overwrite          if output directory for a platform already exists, replaces it rather than skipping it
 prune              runs `npm prune --production` on the app
 sign               should contain the identity to be used when running `codesign` (only for building for the darwin platform, on OS X)
+strict-ssl         whether SSL certificates are required to be valid when downloading Electron (defaults to true)
```
malept: I probably should have caught this first, but how are you supposed to call this on the command line? I think this is the first boolean flag in `electron-packager` that defaults to `true`.

lukeapage: you have to use =false,  minimalist handles it.

malept: I anticipate getting questions about that, so it would be best to put that in `usage.txt`. (Alternatively, the flag should be named `--disable-strict-ssl`.)

lukeapage:
I've expanded the usage to include an example

malept:
:+1: Thanks for your contribution!

lukeapage:
thankyou.. any chance of a release soon?

PR #220, opened 2015-12-13
Don't declare helper app as a protocol handler. Closes #219

### LINKED ISSUE #219: Only set protocol defs on main app, not helper app
cf. this comment: https://github.com/maxogden/electron-packager/commit/a157f716a572515a2d117a8bb51e9da3e34022d3#commitcomment-14948442

I'm going to investigate on my side

fasterthanlime:
cf. #219 for the issue & research :star2:

### DIFF
--- a/mac.js
+++ b/mac.js
@@ -61,7 +61,7 @@ module.exports = {
       }
 
       if (opts.protocols) {
-        helperPlist.CFBundleURLTypes = appPlist.CFBundleURLTypes = opts.protocols.map(function (protocol) {
+        appPlist.CFBundleURLTypes = opts.protocols.map(function (protocol) {
           return {
             CFBundleURLName: protocol.name,
             CFBundleURLSchemes: [].concat(protocol.schemes)


malept:
I'm guessing it's not reasonable to have a test case for this?

Could someone more familiar with how Mac apps work take a look at the PR and associated research that @fasterthanlime did?

fasterthanlime:
> I'm guessing it's not reasonable to have a test case for this?

Well... Travis has OSX machines, and afaict testing the above doesn't require any root privileges but having a fixture mac app that handles protocol properly seems like a lot of work :unamused: 

To me the most important question is why was it coded that way in the beginning, was it tested (even manually, once) to begin with, and does changing this break it for the person who originally added it in ?

fasterthanlime:
I can confirm that this PR fixed it for us (in production, for http://itch.io/app), and I can't wait to switch away from our fork and back to the upstream `electron-packager`.

Not sure who you'd like confirmation from — blame brings up @kfranqueiro for this line, maybe they can help?

kfranqueiro:
I'm pretty sure blame only lists me because I touched that (and a lot in general) when refactoring for what became 5.0.0.  As far as I can tell, this logic even [pre-dates that](https://github.com/maxogden/electron-packager/blob/v4.2.0/mac.js#L62), and it might even pre-date who touched it last at that point too.

Mac bundles are mostly a mystery to me.  Your research sounds plausible but I don't have any other knowledge to confirm/deny it.

malept:
OK, I'm inclined to merge this and wait for any further bug reports to be filed. :smile:

fasterthanlime:
> refactor

Ah, I see.

The research really is pretty straight-forward, even though it was annoying to do, because it involves digging up the OSX commands to reset the right cache.

If both apps are registered as protocol handlers, and you're out of luck, the Helper will be called to handle some of your URLs, and since the Helper really isn't meant to be opened by anything other than the main app itself.. it just doesn't do anything (fails with various negative-integer error codes)

If you're lucky and somehow the main app takes precedence in OSX's cache (wasn't able to pinpoint what exactly determines that..), then things work fine.

As far as I can tell, the helper should never have been registered as protocol handler, if it was tested once and worked it was sheer luck, and that fix fixes it for us and our users (22K mac downloads & counting) 100% of the time.

I'm really not sure what other steps to take to move this issue forward :panda_face:

malept:
> [...] that fix fixes it for us and our users (22K mac downloads & counting) 100% of the time.

Good enough for me.

fasterthanlime:
> OK, I'm inclined to merge this and wait for any further bug reports to be filed.

:dancers: there's already evidence that few people actually used it: I'm the only one pushing for a fix :)

> digging up the OSX commands to reset the right cache.

took me a few minutes of googling to find it again, but for the curious/posterity, the tool in question is lsregister, found at:

```bash
/System/Library/Frameworks/CoreServices.framework/Versions/A/Frameworks/LaunchServices.framework/Versions/A/Support/lsregister
```

You can call it with `-kill` to reset protocol associations — reopening apps directly will start to reassociate protocols.

malept:
Thanks for the bugfix and the associated research!

maxogden:
OK i'll try and cut a new release

maxogden:
ok 5.2.1 is out

fasterthanlime:
:+1: thanks a bunch!

PR #228, opened 2015-12-30
Don't use console.error to print info message "Packaging app for plat…

develar:
In my tests I check process error output and throws error if there is one. In any case I don't see any reason to use here "console.error" — I suppose, it is a typo.

### DIFF
--- a/index.js
+++ b/index.js
@@ -103,7 +103,7 @@ function createSeries (opts, archs, platforms) {
         ]
 
         function createApp (comboOpts) {
-          console.error('Packaging app for platform', platform + ' ' + arch, 'using electron v' + version)
+          console.log('Packaging app for platform', platform + ' ' + arch, 'using electron v' + version)
           series(operations, function () {
             require(supportedPlatforms[platform]).createApp(comboOpts, tmpDir, callback)
           })


malept:
> In my tests I check process error output and throws error if there is one.

Out of curiosity, why not check the exit code? (I assume you're using the CLI.)

> I suppose, it is a typo.

Regarding the PR itself, the current behavior has existed since version 2.0.0 or so. It doesn't appear to be a typo - `console.error` seems to have been used pretty consistently for messages. @maxogden do you happen to remember what the reasoning was behind using `console.error` vs. `console.log`?

maxogden:
Originally I was thinking stdout would be used for a zip output stream etc, so I used stderr for status messages and reserved use of stdout in the future. I don't have strong feelings on whether or not we should be using stdout or stderr for status messages, but making a breaking change now for trivial thing like this probably wouldn't be worth it when you can reliably use the process exit code in a programmatic setting.

develar:
> Out of curiosity, why not check the exit code?
Well, reasonable question. Because it was easy to check that no error output (https://gist.github.com/develar/c274d4f5898e8119bb68)

If you are not going to fix it, well, I leave my current workaround and will check exit code also.

@maxogden IDE (JetBrains products at least) show error output in a different color (red by default) and, so, using "error" instead of "info" is a source of misunderstandings.

malept:
Closing, given Max's comments.

kokarn:
Is this something that never will be revisited or is it open for discussion?

I'm currently building a packaging system for an electron app and a lot of tools use stderr for errors so this is pretty inconsistent :/ 
Right now i'll have to do a dirty check for "Packagin app for platform" and handle that on it's own.

Are you open for a future PR for say 9.* where I switch the .error for .log where applicable?

malept:
I would have to think about whether the zip output stream idea is still feasible given the current state of Electron Packager.

kokarn:
Sure, I'm just trying to bump and old discussion to get definite answers. Thanks for taking the time 🍰

Bosch-Eli-Black:
Would it be possible to revisit this? 🙂

I think it's against convention to output non-errors to STDERR.

In our case, the issue that we ran into was much like @develar's: On our continuous integration server, if a build outputs anything to STDERR, the build is treated as failing. We can work around this, but IMHO, it might be better to change the root behavior.

Thanks 🙂

malept:
The workaround is to set `quiet: true` in the Packager config, which in your case is actually always set.

PR #229, opened 2015-12-30
Convert app-version and build-version to be cross-platform arguments

malept:
`app-version` and `build-version` take precedence over `version-string.ProductVersion` and `version-string.FileVersion`, respectively.

Closes #170. (Well, obsoletes it.)

## Open Question(s)

* [x] Do we get rid of the `version-string` way of setting these values? (and obviously, increment the major version)? In my option, there should only be one (obvious) way to do it. Plus, it's documented in the readme what properties `{app,build}-version` set for the various target OSes.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -214,8 +214,8 @@ If the file extension is omitted, it is auto-completed to the correct extension 
   - `LegalCopyright`
   - `FileDescription`
   - `OriginalFilename`
-  - `FileVersion`
-  - `ProductVersion`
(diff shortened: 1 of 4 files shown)


maxogden:
If its just an extra one liner in the code for backwards compat it's worth it to support the old one, just mark it as deprecated in the docs.

malept:
> If its just an extra one liner in the code for backwards compat it's worth it to support the old one, just mark it as deprecated in the docs.

:+1: Yeah, it's two lines per argument (plus a bunch of lines for tests). I've added the deprecation language to the readme and usage docs.

PR #233, opened 2016-01-08
killall -9 <process name> is not killing the branded-app

kamesh-a:
killall -9 <process name> is not killing the app, because bundle executable still points to 'Electron' Branded apps cannot be killed via app name. Patch fixes that bug.

### DIFF
--- a/mac.js
+++ b/mac.js
@@ -8,11 +8,11 @@ var ncp = require('ncp').ncp
 var series = require('run-series')
 var common = require('./common')
 
-function moveHelpers (frameworksPath, appName, callback) {
-  function rename (basePath, oldName, newName, cb) {
+function rename (basePath, oldName, newName, cb) {
     mv(path.join(basePath, oldName), path.join(basePath, newName), cb)
   }
 
+function moveHelpers (frameworksPath, appName, callback) {
   series([' Helper', ' Helper EH', ' Helper NP'].map(function (suffix) {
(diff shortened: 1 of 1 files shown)


mac.js
```diff
@@ -8,11 +8,11 @@ var ncp = require('ncp').ncp
 var series = require('run-series')
 var common = require('./common')
 
-function moveHelpers (frameworksPath, appName, callback) {
-  function rename (basePath, oldName, newName, cb) {
+function rename (basePath, oldName, newName, cb) {
```
kfranqueiro: You'll need to reduce the indentation of the body of this function since you moved it out of `moveHelpers`.

kamesh-a:
@kfranqueiro  sure.

kamesh-a:
@kfranqueiro Indentation looks fine to me. rename function didn't have any chains.

kfranqueiro:
`rename` (line 11) is at 0 spaces. its body (line 12) is at 4 (should be 2) and its end brace (line 13) is at 2 (should be 0), because the location you moved it from was already inside another function.

malept:
I'm kind of surprised `standard` doesn't check indentation.

malept:
In addition to the style fix, perhaps a test can be added to assert the expected behavior?

malept:
This feature was added in #293 and released in electron-packager 6.0.0.

PR #240, opened 2016-01-19
Make version parameter optional

jiahaog:
The readme states that the if I'm using the programmatic api, the `version` argument is optional.

> version - String
> The Electron version with which the app is built (without the leading 'v') - for example, 0.33.9. See Electron releases for valid versions. If omitted, it will use the version of the nearest local installation of electron-prebuilt.

I realised that if I do not run have a local installation of `electron-prebuilt`, an error will be thrown. Instead of this behaviour, I think it would be simpler for the user if the module makes a query to determine the latest version of electron and use that directly if it is not specified. I've made this implementation in my change, and at the same time, a positive side effect would be that the `--version` command line flag is also an optional parameter.

### DIFF
--- a/index.js
+++ b/index.js
@@ -153,7 +153,7 @@ function createSeries (opts, archs, platforms) {
                 })
               } else {
                 console.error('Skipping ' + platform + ' ' + arch +
-                    ' (output dir already exists, use --overwrite to force)')
+                  ' (output dir already exists, use --overwrite to force)')
                 callback()
               }
             } else {
--- a/index.js
+++ b/index.js
@@ -8,6 +8,7 @@ var mkdirp = require('mkdirp')
 var rimraf = require('rimraf')
 var series = require('run-series')
 var resolve = require('resolve')
+var latestVersion = require('latest-version')
 var getPackageInfo = require('get-package-info')
 var common = require('./common')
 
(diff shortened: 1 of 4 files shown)


maxogden:
@jiahaog Hi and thanks for the PR. I am OK with this, but it would be nice if there were tests for this feature so we could catch regressions in the future in case something breaks this.

index.js
```diff
@@ -152,7 +153,7 @@ function createSeries (opts, archs, platforms) {
                 })
               } else {
                 console.error('Skipping ' + platform + ' ' + arch +
-                  ' (output dir already exists, use --overwrite to force)')
+                    ' (output dir already exists, use --overwrite to force)')
```
malept: Why did this indentation change?

malept:
Apart from the style nit, I have the same comments as Max.

index.js
jiahaog: Oops sorry about that, it must have been something with my IDE that I overlooked

jiahaog:
I'm not entirely sure how should I go about writing a test for this. Based on the test app within the `./test/fixtures/basic`, it's my understanding that the test script executes `npm install` before proceeding to package the app. In this scenario, the automatic inference of the electron version that I've written is bypassed because the packager is able to find a version of the nearest local installation of electron-prebuilt. 

Could you point me in the right direction on how to implement a test?

malept:
You would create a new (barebones) test fixture, without `electron-prebuilt` in the `dependencies` or `devDependencies` of the fixture's `package.json`.

Issue #195, opened 2015-11-14
Rename Binary in OSX (CFBundleExecutable)


FirePanther:
The name of the binary (in Contents > MacOS) is always Electron for me.
In the Info.plist the value of the key CFBundleExecutable is also Electron instead of the name.
The main problem for me is, when I add the app into the Login Items it says Electron:
![Login Items](http://i.dv.tl/avekc.jpg)
If I have multiple Electron apps I can't differentiate them.

imns:
I'm also having the same issue.  It ignores the `<appname>` and names it `Electron.app`.

malept:
@imns there is a PR to fix this (#244) that needs some tests written before we can merge it.

PR #244, opened 2016-01-25
Rename Binary in OSX (CFBundleExecutable) #195

psxjpc:
Tackles the issue #195 by renaming the binary to opts.name

### DIFF
--- a/mac.js
+++ b/mac.js
@@ -49,6 +49,7 @@ module.exports = {
       appPlist.CFBundleDisplayName = opts.name
       appPlist.CFBundleIdentifier = opts['app-bundle-id'] || defaultBundleName
       appPlist.CFBundleName = opts.name
+      appPlist.CFBundleExecutable = opts.name
       helperPlist.CFBundleIdentifier = opts['helper-bundle-id'] || defaultBundleName + '.helper'
       helperPlist.CFBundleName = opts.name
 
@@ -98,6 +99,12 @@ module.exports = {
       }, function (cb) {
         mv(path.dirname(contentsPath), finalAppPath, cb)
       })
+
+	  if (appPlist.CFBundleExecutable !== 'Electron') {
+	      operations.push(function (cb) {
+	        mv(tempPath + '/' + opts.name + '.app/Contents/MacOS/Electron', tempPath + '/' + opts.name + '.app/Contents/MacOS/' + appPlist.CFBundleExecutable, cb)
+	      })
+      }
 
       if (opts.sign) {
         operations.push(function (cb) {


mac.js
```diff
@@ -99,6 +100,12 @@ module.exports = {
         mv(path.dirname(contentsPath), finalAppPath, cb)
       })
 
+	  if (appPlist.CFBundleExecutable !== 'Electron') {
```
malept: In what case will this value be `Electron`? (The bigger question I'm asking is, is the conditional necessary?)

psxjpc: True that :)

malept:
Thank you for your contribution!

Could you add a unit test for your change, to make sure there are no regressions in the future?

Additionally, please note that the continuous integration check failed, in that your change does not conform to our coding standards. That will also need to be fixed before this can be merged.

florian:
Any updates on this? It's a pretty annoying bug

malept:
@florian the following needs to happen in order for this to be merged:

* [ ] Coding standards need to be followed
* [ ] A unit test needs to be added so we can prevent regressions of this behavior in the future
* [ ] The PR needs to be rebased on master

Since this hasn't been updated in over a month, feel free to fork this PR and work on it yourself (provided you follow all of the [pull request guidelines](https://github.com/electron-userland/electron-packager/blob/master/CONTRIBUTING.md#filing-pull-requests) in our documentation.

musocrat:
I've run into a weird bug after these changes. I got it resolved, but can't for the life of me pin down the actual cause. 

I was getting `The application ... can't be opened.` after updating to > 6.0.0. The fix was to change options.name. At first I thought it was due to a period in the name but that wasn't it, as using the same name all lowercased works. Then I thought it was some kind of naming collision with another option, but changing things that were the same like `app-bundle-id` didn't fix it either.

It's an odd thing and I don't expect any further investigation of it as changing the name works. I really just wanted to document it here in case someone else runs into it.

AlicanC:
@musocrat I can confirm. Thanks for writing the post, it saved me a good amount of time :+1:

develar:
@musocrat Not quite understand — what was changed to fix? If name should be somehow normalised — it should be done by tool :)

AlicanC:
@develar, the _fix_ is using "my app name" instead of "My App Name". Capitalized name used to work on 5.1.0, but they got broken when I updated to 6.0.2 today.

develar:
@AlicanC Thanks, it will be investigated and fixed in the electron-builder as part of https://github.com/electron-userland/electron-builder/issues/288

PR #247, opened 2016-01-26
Allow ignore to be a Function

paulcbetts:
This PR allows the `ignore` option to be a Function as well as a Regexp

### DIFF
--- a/test/basic.js
+++ b/test/basic.js
@@ -346,6 +346,7 @@ util.testAllPlatforms('prune test', createPruneTest)
 util.testAllPlatforms('ignore test: string in array', createIgnoreTest, ['ignorethis'], 'ignorethis.txt')
 util.testAllPlatforms('ignore test: string', createIgnoreTest, 'ignorethis', 'ignorethis.txt')
 util.testAllPlatforms('ignore test: RegExp', createIgnoreTest, /ignorethis/, 'ignorethis.txt')
+util.testAllPlatforms('ignore test: Function', createIgnoreTest, function (file) { return file.match(/ignorethis/) }, 'ignorethis.txt')
 util.testAllPlatforms('ignore test: string with slash', createIgnoreTest, 'ignore/this',
   path.join('ignore', 'this.txt'))
 util.testAllPlatforms('ignore test: only match subfolder of app', createIgnoreTest, 'electron-packager',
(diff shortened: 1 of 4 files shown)


test/basic.js
```diff
@@ -346,6 +346,7 @@ util.testAllPlatforms('prune test', createPruneTest)
 util.testAllPlatforms('ignore test: string in array', createIgnoreTest, ['ignorethis'], 'ignorethis.txt')
 util.testAllPlatforms('ignore test: string', createIgnoreTest, 'ignorethis', 'ignorethis.txt')
 util.testAllPlatforms('ignore test: RegExp', createIgnoreTest, /ignorethis/, 'ignorethis.txt')
+util.testAllPlatforms('ignore test: RegExp', createIgnoreTest, function (file) { return !(file.match(/ignorethis/)) }, 'ignorethis.txt')
```
malept: This should probably be named differently.

paulcbetts: hwhoops, agree

malept:
Thanks for the PR! I have no objections to this new feature, does any other contributor have any constructive criticism?

(@paulcbetts, once there's a consensus, could you please squash your commits?)

kfranqueiro:
My one concern is that specifying `ignore` as a function, but then requiring it to return `false` to *ignore*, feels kind of backwards.  (This is primarily awkward because previously the function was internal whereas now it's directly exposed.)  I realize the doc update explains that, but I figure it's still apt to throw people.

Given that the code calls the function within the function it actually passes to `filter` anyway, could we throw `!` in front of that call so that `ignore` as a function could be implemented to return `true` to ignore instead?

kfranqueiro:
Also, it seems as if the added test failed on CI?

paulcbetts:
> (@paulcbetts, once there's a consensus, could you please squash your commits?)

nbd

> Given that the code calls the ignore function within the function it actually passes to filter anyway, could we throw ! in front of that call so that ignore as a function could be implemented to return true to ignore instead?

I'm okay with that

> Also, it seems as if the added test failed on CI?

Yeah, I gotta fix that up

maxogden:
lgtm! only nit is that `name` is a reserved keyword so I normally don't use it, but it probably doesn't matter here

Issue #64, opened 2015-06-13
Windows Build ENOENT Error.


delvedor:
Hi, I'm trying to build my app for windows from a mac.
These are my scripts, the Electron-packager version is 4.1.2:
```javascript
"build-mac": "electron-packager ./ MyApp --platform=darwin --arch=x64 --version=0.28.1"
"build-win": "electron-packager ./ MyApp --platform=win32 --arch=x64 --version=0.28.1"
"build-win": "electron-packager ./ MyApp --platform=win32 --arch=ia32 --version=0.28.1"
```
The build for mac goes right, contrariwise both the builds for Windows gave me this error (just change the arch with x64, I modified slightly the output by masking directory paths):
```
Geyser:MyApp delvedor$ npm run-script build-win

> myapp@3.1.0 build-win /.../.../MyApp
> electron-packager ./ MyApp --platform=win32 --arch=ia32 --version=0.28.1

Downloading electron-v0.28.1-win32-ia32.zip
[=================================================>] 100.0% of 38.79 MB (1.12 MB/s)
Packaging app for platform win32 ia32 using electron v0.28.1
[ { [Error: ENOENT, stat '/.../.../MyApp.app/Contents/Frameworks/Electron Framework.framework/Libraries/Libraries']
    errno: -2,
    code: 'ENOENT',
    path: '/.../.../MyApp.app/Contents/Frameworks/Electron Framework.framework/Libraries/Libraries' },
  { [Error: ENOENT, stat '/.../.../MyApp.app/Contents/Frameworks/Electron Framework.framework/Versions/Current/Libraries/Libraries']
    errno: -2,
    code: 'ENOENT',
    path: '/.../.../MyApp.app/Contents/Frameworks/Electron Framework.framework/Versions/Current/Libraries/Libraries' },
  { [Error: ENOENT, stat '/.../.../MyApp.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Libraries/Libraries']
    errno: -2,
    code: 'ENOENT',
    path: '/.../.../MyApp.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Libraries/Libraries' } ] undefined

npm ERR! Darwin 14.3.0
npm ERR! argv "node" "/usr/local/bin/npm" "run-script" "build-win"
npm ERR! node v0.12.0
npm ERR! npm  v2.7.6
npm ERR! code ELIFECYCLE
[...]

PR #223, opened 2015-12-20
Platform: Mac App Store support

sethlu:
- MAS platform support
- Entitlements (optional) are supported to sign apps, not just with a certificate
- Made it easier to sign for MAS distribution, may fallback to default entitlements file

PR #255, opened 2016-02-05
Automatically ignore --out path

develar:
This pull request supersedes #155. And is required for #251.

Tests and doc added.

### DIFF
--- a/common.js
+++ b/common.js
@@ -50,8 +50,19 @@ function userIgnoreFilter (opts) {
   }
 
   var normalizedOut = opts.out ? path.resolve(opts.out) : null
+  var outIgnores = []
+  if (normalizedOut === null || normalizedOut === process.cwd()) {
+    ['darwin', 'linux', 'win32'].forEach(function (platform) {
+      ['ia32', 'x64'].forEach(function (arch) {
+        outIgnores.push(path.join(process.cwd(), opts.name + '-' + platform + '-' + arch))
+      })
+    })
+  } else {
+    outIgnores.push(normalizedOut)
+  }
+
   return function filter (file) {
-    if (normalizedOut === file) {
+    if (outIgnores.indexOf(file) !== -1) {
       return false
     }
 
--- a/common.js
+++ b/common.js
@@ -49,7 +49,12 @@ function userIgnoreFilter (opts) {
     }
   }
 
+  var normalizedOut = opts.out ? path.resolve(opts.out) : null
   return function filter (file) {
+    if (normalizedOut === file) {
+      return false
+    }
+
     var name = file.split(path.resolve(opts.dir))[1]
 
     if (path.sep === '\\') {
--- a/common.js
+++ b/common.js
@@ -9,6 +9,9 @@ var ncp = require('ncp').ncp
 var rimraf = require('rimraf')
 var series = require('run-series')
 
+var archs = ['ia32', 'x64']
+var platforms = ['darwin', 'linux', 'mas', 'win32']
+
 function asarApp (appPath, asarOptions, cb) {
   var src = path.join(appPath)
   var dest = path.join(appPath, '..', 'app.asar')
@@ -52,8 +55,8 @@ function userIgnoreFilter (opts) {
   var normalizedOut = opts.out ? path.resolve(opts.out) : null
   var outIgnores = []
   if (normalizedOut === null || normalizedOut === process.cwd()) {
-    ['darwin', 'linux', 'win32'].forEach(function (platform) {
-      ['ia32', 'x64'].forEach(function (arch) {
+    platforms.forEach(function (platform) {
+      archs.forEach(function (arch) {
         outIgnores.push(path.join(process.cwd(), opts.name + '-' + platform + '-' + arch))
       })
     })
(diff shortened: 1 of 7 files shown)


malept:
I would prefer that since you're building on #155, that you rebase + squash that branch so that the original author is still attributed to their changes, and then an additional commit is for you adding the tests requested.

develar:
Sorry for noise, I created pull request too early (realized, that my test is not correct). I will add comment when it will be ready. Yeah, writing tests is not easy.

develar:
Strange. Tests passed locally, but not on Travis. And in your build (https://travis-ci.org/maxogden/electron-packager/jobs/107157456) I see the same error. I will start build again in a few hours.

develar:
> so that the original author is still attributed to their changes

Fixed, I set author to `pyro2927 <joseph@pintozzi.com>`.

develar:
I don't see what's wrong in my code. And Travis build passed. (I see that another unrelated build also blinking).

Maybe, it will be better to test such functionality only for current platform (not for all platforms) to speedup tests, but I used standard `util.testAllPlatforms`.

malept:
Could you add one more test for the new ignore functionality when you don't specify an `out`? (That's what I meant in the other PR for a "current working directory case".

develar:
> Could you add one more test for the new ignore functionality when you don't specify an out

It is required to fix #64 I don't like it, but as a tool we cannot teach users, ok. Current solution is not complete, thanks for pointing.

common.js
```diff
@@ -49,7 +49,23 @@ function userIgnoreFilter (opts) {
     }
   }
 
+  var normalizedOut = opts.out ? path.resolve(opts.out) : null
+  var outIgnores = []
+  if (normalizedOut === null || normalizedOut === process.cwd()) {
+    ['darwin', 'linux', 'win32'].forEach(function (platform) {
+      ['ia32', 'x64'].forEach(function (arch) {
```
malept: I would move `supportedArchs` and `supportedPlatforms` in `index.js` to `common.js` and use those instead. Otherwise, that will make PRs like #107 and #223 more complex than they need to be.

develar:
Tests failed due to critical error in all tests — 
1) you pass `combination` to test and then call `Object.create`.
2) Created opts has `combination` as a prototype.
3) `createInferTest` calls `delete combination.version` directly on original `combination` instance and... yes, `version` deleted from prototype and all tests now could be broken.

As part of this pull request I will fix it.

develar:
> I would move supportedArchs and supportedPlatforms in index.js to common.js
Sorry, tried to minimise, but it is stupid, yes. Fixed.

index.js
```diff
@@ -11,17 +11,16 @@ var resolve = require('resolve')
...
+var supportedArchs = common.archs.reduce(function (o, v) {
+  o[v] = 1
+  return o
+}, {})
+
+// Maps to module ID for each platform (lazy-required if used)
+var supportedPlatforms = common.platforms.reduce(function (o, v) {
+  o[v] = './' + (v === 'darwin' ? 'mac' : v)
+  return o
+}, {})
```
malept: In my opinion, this makes it a bit more difficult to understand what's going on.

develar: I also prefer to use `for of` instead of magic `function style`. But I see `forEach` in existing code.

Do you want to convert it to

```js
var supportedPlatforms = {}
for (var platform in common.platforms) {
  supportedPlatforms[platform] = './' + (platform === 'darwin' ? 'mac' : platform)
}
```
?

malept: At the very least, please don't use single-letter variables here. Please also keep in mind that there is a PR to add the Mac App Store (MAS) platform (#223) which currently also uses the `mac.js` file.

develar: Variables renamed. I don't want to pollute file scope, but cannot use `let`, so, still use `reduce` instead of `for in` (`for of` es6).

Mac App Store (MAS) platform can be added as before — directly to supportedPlatforms (or inside reduce function — in any case we check `is equal to darwin?`).

malept: I think with your current code, the MAS change will look like this:

```javascript
result[platform] = './' + (['darwin', 'mas'].indexOf(platform) !== -1 ? 'mac' : platform)
```

It's not great.

develar: Why? Simple line after:
`supportedPlatforms.mas = './mac'`

develar: Or 

```
 var supportedPlatforms = common.platforms.reduce(function (result, platform) {
   result[platform] = './' + (platform === 'darwin' ? 'mac' : platform)
   return result
 }, {
  mas: './mac'
})
```

malept: Seems like you may as well do either option for `darwin` as well and get rid of the ternary operator. (Though I'm not convinced that the second option actually works.)

develar: It seems, you don't like  ternary operator :)

Ok, my code saves only 2 lines (linux and windows). Maybe we can rename mac.js to darwin.js to consolidate, but I don't think that it is an acceptable solution.

So, I reverted this block to previous.

malept: I'm OK with ternary operators when all parts of the statement are simple.

develar: Travis build passed, please review.

develar:
Should I fix something or you just haven't yet time to review it?

malept:
Sorry, I've been holding off reviewing/merging other PRs because #223 is bigger than the others. Hopefully in the next couple of days.

malept:
Yep, I figured that you would need to rebase after I merged the PR I mentioned earlier.

develar:
@malept I missed your quick reply, sorry. I have rebased.

Issue #235, opened 2016-01-10
Not currently possible to ignore devDependencies with npm3


maxogden:
If you are packaging an app and using npm3, there is no way to easily ignore devDependencies because [npm3 flattens dependencies](https://docs.npmjs.com/how-npm-works/npm3), so there is no way to know ahead of time what things are necessary to include in the `--ignore` list.

I just wanted to mention it here in case someone runs into this issue. The fix is to use the `--yes-prune` flag. Maybe we should consider defaulting this to `true` in the next major version? Is there a downside (other than backwards compat) to defaulting it to true?

malept:
> Maybe we should consider defaulting this to `true` in the next major version?

No objections from me.

We should probably start keeping a `changelog.md` (or `news.md`) for these sorts of breaking changes. (See: http://keepachangelog.com/)

malept:
(See also: part of #35)

PR #263, opened 2016-02-14
Add NEWS.md, a human-readable list of changes

malept:
As mentioned in https://github.com/maxogden/electron-packager/issues/235#issuecomment-170384955 and https://github.com/maxogden/electron-packager/pull/223#issuecomment-183989800.

Only contains changes from 5.2.0 onward. If someone else wants to backfill, feel free :grin:

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,0 +1,59 @@
+# Changes by Version
+
+## Unreleased
+
+### Added
+
+* Add support for a new target platform, Mac App Store (`mas`) (#223)
+* Add `app-copyright` parameter (#223)
+* Add `tmpdir` parameter to specify a custom temp directory (#230)
+* Add `NEWS.md`, a human-readable list of changes in each version (since 5.2.0) (#263)
+
+### Changed
+
+* Allow the `ignore` parameter to take a function (#247)
+* [contributors] Update Standard (JavaScript coding standard) package to 5.4.x
+* [contributors] Add code coverage support via Coveralls (#257)
+* Better docs around contributing to the project (#258)
+
+### Fixed
+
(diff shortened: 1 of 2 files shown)


NEWS.md
```diff
@@ -0,0 +1,53 @@
+# Changes by Version
+
+## [Unreleased]
+
+### Added
+
+* Add `tmpdir` parameter to specify a custom temp directory (#230)
```
malept: Opinions on whether to actually link the PRs? I kind of wish GitHub would automatically do so.

kfranqueiro: If we were to use GitHub's releases feature instead of add a file to the repo ourselves, it automatically links issues there.

malept: I propose we do both. This way the person tagging the release is not on the hook to gather the release notes, it should be gathered as changes are added.

kfranqueiro: FWIW, any contributor is able to create or edit a release draft on github at any time to keep track of release notes ahead of an actual release; e.g. with dgrid I try to keep a draft updated as each thing gets merged. Just have to be careful not to prematurely click the big green button on the page (why do they have to always have tempting big green buttons?!)

Not sure if that influences your thought of doing both a NEWS.md and github release notes.

malept: > Just have to be careful not to prematurely click the big green button on the page (why do they have to always have tempting big green buttons?!)

Yeah, today I already accidentally clicked the "close PR" button.

The nice thing about the `NEWS.md` approach is that a user can tell what's unreleased - using just GitHub Releases doesn't really help with that use case. I have needed to figure out if feature X has been released or not when responding to issues.

malept:
I added GitHub compare links to the version names, but for whatever reason, the diffs aren't quite right.

PR #274, opened 2016-02-29
Split up arguments section into OS-specific subsections

### LINKED ISSUE #237: Split up arguments section into OS-specific subsections
From https://github.com/maxogden/electron-packager/issues/197#issuecomment-161055204:

> The optional arguments section should be divided into "all platforms" / "Windows Only" / "OS X only" (and then alphabetized within each subsection).

malept:
Fixes #237.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -146,21 +146,11 @@ packager(opts, function done (err, appPath) { })
 
 **Optional**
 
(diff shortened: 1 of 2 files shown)


malept:
Blocked by #278.

Issue #275, opened 2016-02-29
Reorganize docs


malept:
I think the readme has gotten way too long. As of #274, I think it's at seven levels of headers. The API reference should go in a separate document.

PR #276, opened 2016-02-29
Ability to set tmpdir to false to disable using temp directory

develar:
Continue #251.

Test added.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -22,6 +22,11 @@ if (!args.dir || (!args.all && (!args.platform || !args.arch))) {
   process.exit(1)
 }
 
+// minimist doesn't support multiple types for a single argument (in this case, `String` or `false`)
+if (args.tmpdir === 'false') {
+  args.tmpdir = false
+}
+
 packager(args, function done (err, appPaths) {
   if (err) {
     if (err.message) console.error(err.message)
--- a/common.js
+++ b/common.js
@@ -1,11 +1,9 @@
 var child = require('child_process')
-var fs = require('fs')
+var fs = require('fs-extra')
 var os = require('os')
 var path = require('path')
 
 var asar = require('asar')
-var mkdirp = require('mkdirp')
-var ncp = require('ncp').ncp
 var rimraf = require('rimraf')
 var series = require('run-series')
 
@@ -94,27 +92,23 @@ module.exports = {
     // * Prunes non-production node_modules (if opts.prune is set)
     // * Creates an asar (if opts.asar is set)
 
-    var tempParent = path.join(opts.tmpdir || os.tmpdir(), 'electron-packager', opts.platform + '-' + opts.arch)
-    var tempPath = path.join(tempParent, generateFinalBasename(opts))
+    var tempPath
+    if (opts.tmpdir === false) {
+      tempPath = generateFinalPath(opts)
+    } else {
+      tempPath = path.join(opts.tmpdir || os.tmpdir(), 'electron-packager', opts.platform + '-' + opts.arch, generateFinalBasename(opts))
+    }
+
     // Path to `app` directory
     var appPath = path.join(tempPath, appRelativePath)
     var resourcesPath = path.resolve(appPath, '..')
 
     var operations = [
       function (cb) {
-        rimraf(tempParent, function () {
-          // Ignore errors (e.g. directory didn't exist anyway)
-          cb()
-        })
+        fs.move(templatePath, tempPath, {clobber: true}, cb)
       },
       function (cb) {
-        mkdirp(tempPath, cb)
-      },
-      function (cb) {
-        ncp(templatePath, tempPath, cb)
-      },
-      function (cb) {
-        ncp(opts.dir, appPath, {filter: userIgnoreFilter(opts), dereference: true}, cb)
+        fs.copy(opts.dir, appPath, {filter: userIgnoreFilter(opts), dereference: true}, cb)
       },
       function (cb) {
         rimraf(path.join(resourcesPath, 'default_app'), cb)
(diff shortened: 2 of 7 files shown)


common.js
```diff
@@ -151,15 +145,13 @@ module.exports = {
...
-        ncp(tempPath, finalPath, cb)
-      }
-    ], function (err) {
+
+    if (opts.tmpdir === false) {
+      callback(null, finalPath)
+      return
+    }
+
+    fs.move(tempPath, finalPath, function (err) {
```
develar: fs-extra handles `EPERM` on windows [correctly](https://github.com/jprichardson/node-fs-extra/blob/master/lib/move/index.js). We don't need to use mkdirp/ncp here explicitly.

malept:
Is this ready for review now? I don't want to merge this before you're ready (again).

develar:
I suppose #251 was closed automatically because was rebased to the same commits as #255) by GitHub, not by you.

Yes, please review.

cli.js
```diff
@@ -22,6 +22,10 @@ if (!args.dir || (!args.all && (!args.platform || !args.arch))) {
   process.exit(1)
 }
 
+if (args.tmpdir === 'false') {
+  args.tmpdir = false
+}
+
```
malept: Should add a comment that minimist doesn't do this for us. I almost forgot why this conditional was added.

index.js
```diff
@@ -124,30 +127,32 @@ function createSeries (opts, archs, platforms) {
...
-          function (cb) {
-            mkdirp(tmpDir, cb)
-          },
-          function (cb) {
-            extract(zipPath, {dir: tmpDir}, cb)
-          }
-        ]
-
         function createApp (comboOpts) {
+          var buildDir = path.join(useTempDir ? tempBase : (opts.out || process.cwd()), platform + '-' + arch + '-template')
```
malept: Could you extract the ternary statement into a separate variable + if/else statement? It makes this line easier to read.

cli.js
malept: Please also add documentation to `readme.md` and `usage.txt` about this new functionality to `tmpdir`. Though you may want to hold off on that until #275 is merged.

cli.js
```diff
@@ -22,6 +22,11 @@ if (!args.dir || (!args.all && (!args.platform || !args.arch))) {
   process.exit(1)
 }
 
+// minimist doesn't support to specify several types (string|boolean)
```
develar: Commend added.

common.js
```diff
@@ -1,11 +1,9 @@
 var child = require('child_process')
-var fs = require('fs')
+var fs = require('fs-extra')
 var os = require('os')
 var path = require('path')
 
 var asar = require('asar')
-var mkdirp = require('mkdirp')
```
develar: Using fs-extra was approved on first iteration (no negative comments), so, I have decided to cleanup — no need to to use mkdirp/rimraf explicitly if we use fs-exra.

index.js
```diff
@@ -1,11 +1,9 @@
 var path = require('path')
-var fs = require('fs')
+var fs = require('fs-extra')
 var os = require('os')
 
 var download = require('electron-download')
 var extract = require('extract-zip')
-var mkdirp = require('mkdirp')
```
develar: Using fs-extra was approved on first iteration (no negative comments), so, I have decided to cleanup — no need to to use mkdirp/rimraf explicitly if we use fs-exra

cli.js
malept: It would read better as:

```javascript
// minimist doesn't support multiple types for a single argument (in this case, `String` or `false`)
```

develar: Fixed, thanks.

readme.md
```diff
@@ -248,7 +248,7 @@ If the file extension is omitted, it is auto-completed to the correct extension
 
 `tmpdir` - *String*
 
-  The base directory to use as a temp directory. Defaults to the system temp directory.
+  The base directory to use as a temp directory. Defaults to the system temp directory. Set to `false` to write to `out` directly without temp.
```
develar: Docs added.

usage.txt
```diff
@@ -33,7 +33,7 @@ sign               should contain the identity to be used when running `codesign
 sign-entitlements  the path to entitlements used in signing (mas platform only)
 strict-ssl         whether SSL certificates are required to be valid when downloading Electron.
                    It defaults to true, use --strict-ssl=false to disable checks.
-tmpdir             temp directory. Defaults to system temp directory.
+tmpdir             temp directory. Defaults to system temp directory. Set to `false` to write to `out` directly without temp.
```
malept: For consistency:

> Defaults to system temp directory, use --tmpdir=false to disable functionality.

readme.md
```diff
@@ -248,7 +248,7 @@ If the file extension is omitted, it is auto-completed to the correct extension
 
 `tmpdir` - *String*
```
malept: >  `tmpdir` - *String* or *false*

usage.txt
develar: Changed, but don't quite agree that "to disable functionality." is clear — because result of "disable functionality." is not clear. Opposite to "to write to `out` directly without temp".

malept: Perhaps:

> Specify --tmpdir=false to disable use of a temporary directory.

develar: I feel a little stupid compared to you :) Changed.

malept: Well, I was trying to avoid having it say some variant of "temp directory" three times in three sentences - I guess that didn't work out. (I am open to suggestions.)

malept:
:+1:

Issue #271, opened 2016-02-21
Entitlements for saving to user-selected dir


positlabs:
I am trying to save a user-selected output directory path so I can save to it in the future, without having the user re-select the dir every time the app is restarted. I think I have more than enough entitlements in the plists, so I'm not sure why it's borking.

> Which version of electron-packager are you using?

5.2.1 (but actually latest/master)

> What parameters are you passing to the `packager()` function?

```
{
		dir: './',
		name: 'GifGoat',
		'app-bundle-id': 'GIFGOAT-HELPER',
		icon: './assets/icons/gifgoat-icon.icns',
		// platform: 'darwin',
		platform: 'mas',
		arch: 'x64',

		version: require('./node_modules/electron-prebuilt/package.json').version, // v0.36.8
		'app-version': require('./package.json').version,
		'build-version': require('./package.json').version,
		
		asar: true,
		'asar-unpack-dir': 'bin/',

		'app-bundle-id': '...',
		'app-category-type': 'public.app-category.utilities',

		sign: '3rd Party Mac Developer Application: ...',
		'sign-entitlements': './build/mac-extras/parent.plist',
		'entitlements-inherit': './build/mac-extras/child.plist',
		prune: true,
		ignore: [
			'/bin/ffmpeg/linux($|/)',
			'/bin/ffmpeg/win32($|/)',
			'/dist($|/)',
			'/build($|/)',
			'/test($|/)',
		],
		out: './dist/',
		overwrite: true
}
```

> What is the host platform are you running electron-packager on?

OSX 10.11.3

> What target platform(s)/architecture(s) are you building for?

darwin & mas, x64

> Is there a stack trace in the error message you're seeing?

<img width="686" alt="screen shot 2016-02-21 at 3 03 46 pm" src="https://cloud.githubusercontent.com/assets/1918844/13206092/66172e6a-d8ac-11e5-89dd-358de7896693.png">

> Please provide either a failing testcase or detailed steps to reproduce your problem.

parent.plist
```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
  <dict>
    <key>com.apple.security.app-sandbox</key>
    <true/>
    <key>com.apple.security.application-groups</key>
    <true/>
    <key>com.apple.security.files.user-selected.read-write</key>
    <true/>
[...]

PR #278, opened 2016-02-29
Add option for child entitlements

mcfedr:
Allows the `entitlements-inherit` option of `electron-osx-sign` to be passed.

Related to #271

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -240,7 +240,11 @@ If the file extension is omitted, it is auto-completed to the correct extension 
 
 `sign-entitlements` - *String*
 
-  The path to entitlements used in signing. (Currently limited to Mac App Store distribution.)
+  The path to the 'parent' entitlements used in signing.
+
+ `sign-entitlements-inherit` - *String*
+
+  The path to the 'child' entitlements. See [electron-osx-sign](https://www.npmjs.com/package/electron-osx-sign#opts) for more details.
 
 `strict-ssl` - *Boolean*
 
--- a/mac.js
+++ b/mac.js
@@ -163,7 +163,8 @@ module.exports = {
             // as boolean value true. Then fallback to null for auto discovery,
             // otherwise provided signing certificate.
             identity: opts.sign === true ? null : opts.sign,
-            entitlements: opts['sign-entitlements']
+            entitlements: opts['sign-entitlements'],
+            'entitlements-inherit': opts['sign-entitlements-inherit']
           }, function (err) {
             if (err) {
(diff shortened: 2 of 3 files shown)


malept:
Thanks for the contribution! Could you please add an entry to `usage.txt`?

mcfedr:
I didn't re-indent the whole file right away, it pushes everything very far to the right, what do you think @malept ?

malept:
Hmm. Try moving the description to the next line.

Also, un-capitalize "The", it's inconsistent with the rest of the argument descriptions.

mcfedr:
ah, I even looked at the capitals and thought it should be small.

readme.md
```diff
@@ -240,7 +240,11 @@ If the file extension is omitted, it is auto-completed to the correct extension
 
 `sign-entitlements` - *String*
 
-  The path to entitlements used in signing. (Currently limited to Mac App Store distribution.)
+  The path to the 'parent' entitlements used in signing. (Currently limited to Mac App Store distribution.)
+  
+ `sign-entitlements-inherit` - *String*
+ 
+  The path to the 'child' entitlements. See [electron-osx-sign](https://www.npmjs.com/package/electron-osx-sign) for more details.
```
malept: This is a more accurate link: https://www.npmjs.com/package/electron-osx-sign#opts

malept: Also, you should add the same caveat as `sign-entitlements`:

> (Currently limited to Mac App Store distribution.)

sethlu: @malept The signing with entitlements should now work fine with `darwin` builds in the latest `electron-osx-sign v0.3.0`.

malept: @sethlu thanks for the correction. I guess the caveat needs to be removed from the `entitlements` docs, then. That can happen outside of this PR.

sethlu: @malept That may be better I think.
I'll find a way to explain the entitlements for `darwin` builds as they, apart from sandboxing an app, do not actually do much except interacting with iCloud services, iAP or things like that. It's pretty hard to access some of these Apple services with current builds of Electron, so basically entitlements could be ignored for `darwin`. The default entitlements for `darwin` is blank: If a `darwin` build is sandboxed, it should not also really function properly in the presence of frameworks like `Squirrel.framework`.

malept:
Looks good. Squash the commits and I'll merge.

mcfedr:
@malept squashed

malept:
Thanks!

PR #280, opened 2016-03-02
Update osx sign

mcfedr:
The new version solves a number of bugs where dylibs are not all signed

### DIFF
--- a/mac.js
+++ b/mac.js
@@ -166,7 +166,7 @@ module.exports = {
             entitlements: opts['sign-entitlements']
           }, function (err) {
             if (err) {
-              console.warn('Code sign failed; please retry manually.')
+              console.warn('Code sign failed; please retry manually.', err)
               // Though not signed successfully, the application is packed.
               // It might have to be signed for another time manually.
             }
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
   "dependencies": {
     "asar": "^0.8.2",
     "electron-download": "^1.0.0",
-    "electron-osx-sign": "^0.1.6",
+    "electron-osx-sign": "^0.3.0",
     "extract-zip": "^1.0.3",
     "fs-extra": "^0.26.5",
     "get-package-info": "0.0.2",


malept:
I'm not sure we want to depend on a beta version of electron-osx-sign.

CC: @sethlu

mcfedr:
I know what you mean, but its the only version that actually does the job

positlabs:
You could always sign manually, as described here: http://electron.atom.io/docs/v0.36.8/tutorial/mac-app-store-submission-guide/#sign-your-app

malept:
Additionally, according to Travis, upgrading to this version breaks compatibility with Node 0.12.

mcfedr:
Its true I could sign seperately, but the point was that this tool tries to solve the whole process, and currently it doesnt because its using an old version of `electron-osx-sign` that is broken.

I hadn't see that it break Node 0.12 compatibility, my old comment would be that for a build tool that is unlikely to be a problem, its not like people are trying to deploy to 'stable' server environment, I doubt its a problem for any build environments to use a more up to date version of Node.

malept:
You'd be surprised how infrequently build servers get updated (unless there's a security fix required).

mcfedr:
I would have thought anyone using electron, which has only been around for a short time, is using a relatively new server.
Also, if `electron-packager` is unable to sign an app, you still have to use `electron-osx-sign` so the newer node dependancy still remains.

malept:
My point in mentioning it was to get @sethlu's comment.

sethlu:
@malept I'll be releasing `electron-osx-sign v3.0` later today after adding the signing with entitlements for `darwin` builds with default entitlements file following @jasonhinkle's comment on https://github.com/electron-userland/electron-osx-sign/issues/15 yesterday, and some extra manual tests. The only limitation with the current beta is that some binaries are not detected. This should be resolved with the release that'll be out very soon.

malept:
@mcfedr please update with new non-beta version of electron-osx-sign (which also fixes the Node version problem I mentioned).

malept:
Looks good. Please squash the commits (per [the pull request instructions](https://github.com/electron-userland/electron-packager/blob/master/CONTRIBUTING.md#filing-pull-requests)) and I'll merge.

Issue #285, opened 2016-03-09
Pass binaries option to electron-osx-sign


positlabs:
I have some extra binaries that need signing. Currently, I'm signing them with a bash script, but ideally it would be all handled by `electron-packager`. 

Since this is already a feature of `electron-osx-sign`, it could be solved by adding an extra argument; `binaries` or maybe `osx-binaries` for clarity. This argument would then be passed to `electron-osx-sign`.

Let me know what you think and I'll create a PR.

malept:
I don't have any reason to be against the feature in general. I'd call it `osx-binaries`.

However, I'm beginning to think there needs to be a better way to pass arguments to `electron-osx-sign`. That would be the fourth argument that just passes information to that module. Perhaps it should be named `sign-options`, and function like how `version-string` works for Windows targets. I would even be willing to delay releasing 6.0.0 (that I was going to release today) if it meant saner parameters moving forward.

CC: @sethlu

positlabs:
Agreed. Makes sense to group them. 

I'm imagining something like this:
```
'osx-sign': {
  'bundle-identifier': {String},
  'entitlements': {path},
  'entitlements-inherit': {path},
  'binaries': {Array of paths}
}
```

malept:
That seems reasonable - then for the hypothetical Windows version, you would have `windows-sign`.

If I were to work on this, I would reorganize the docs first, then implement.

positlabs:
I'm pretty busy this week, but I can pick it up this weekend.

sethlu:
It's a good point to group those options; however, we may need some extra lines of code for this to work with the command line as we are trying to put several arguments into one object like `electron-packager ... --osx-sign`? But this looks like a neat solution to cross-platform support.

I'm thinking about the integration of `opts.all`: If we're separating the different platforms' opts then someone using it should specify for each of the platforms the different params. Would it be better just to attach some prefixes so they still work with the command line arguments while keeping from nesting objects in `opts`? And then docs could be categorized for the various platforms.

@jasonhinkle and I have briefly discussed in https://github.com/electron-userland/electron-osx-sign/issues/13 on signing on the Windows platform.

malept:
@sethlu here is an example of a set of options that are grouped together already, that's Windows-specific: https://github.com/electron-userland/electron-packager/blob/c2dd6c26c19223fff6e1ef0a411e3a7ce8e94ffe/usage.txt#L48-L59

malept:
[...]

Issue #32, opened 2015-05-12
Add --sign to Windows


jden:
see #31 

It would be awesome to be able to output signed binaries.

I'm not positive if it's possible to sign executables after compilation with the current rcedit strategy we use. TODO: investigate this.

I'm actually pretty unfamiliar with the Visual Studio code signing tools overall. If anyone else knows what they're doing, feel free to pitch in! Otherwise, I'll try to investigate when I get time.

Rameshv:
I have too many issues with https://github.com/atom/grunt-electron-installer for creating installers and code signing. I can't get this even compile in my local/build servers. 

So i went ahead and implemented my own app installer creation (using inno setup https://github.com/jrsoftware/issrc instead the atom squirrel) and code signing with signtool.. it works perfect... its currently a bit messy. 

Its currently set of separate grunt tasks.  If you are looking to go this way, let me know..  i ll integrate with electron-package (along with this https://github.com/maxogden/electron-packager/issues/30) and make a pull request soon

jden:
@Rameshv do you have this work in a repo somewhere? We'll be happy to work with you to review it and integrate it with electron-packager as cleanly as possible. We're currently not using grunt, and I'd like to avoid introducing large new dependencies like that if we don't need to.

coclav:
I am using electron-boilerplate but I trust the logic would be the same here. I sign my app twice. First the "app.exe" and then the "app_installer.exe" 

```
var sign = childProcess.spawn(resourcesDir.path("signtool"),
	[
		'sign',
		'/f',
		resourcesDir.path("MyCertificate.p12"),
		'/p',
		'APassword',
		'/t',
		'http://timestamp.verisign.com/scripts/timstamp.dll',
		readyAppDir.path(manifest.productName + '.exe')
	],
	{
    		stdio: 'inherit'
  	});

  sign.on('error', function (err) {
      if (err) {
          throw "sign error " + err;
      } else {
          throw err;
      }
  });
  sign.on('close', function () {
      gulpUtil.log('Sign done!');
      deferred.resolve();
  });
```

And then this is the "packaged" into the installer, that I will sign the same way.

bouncyball02:
According to [the Electron documentation](https://www.electronjs.org/docs/tutorial/code-signing#signing-windows-builds), some tools already offer this.

Specifically:
- [electron-winstaller](https://github.com/electron/windows-installer)
- [electron-forge](https://github.com/electron-userland/electron-forge)
[...]

PR #286, opened 2016-03-10
Consolidate OSX signing params

malept:
Per discussion in #285, I feel that there are too many OS X signing options, and that it would be better to group them together. This has the added benefit of leaving in room for a potential future Windows signing module to be integrated (#32).

CC: @sethlu

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -27,6 +27,9 @@
   files (#253)
 * Split out the code to sign OS X apps into a separate Node module,
   [electron-osx-sign](https://github.com/electron-userland/electron-osx-sign) (#223)
+* [darwin/mas] **BREAKING**: The `sign` parameter is now `osx-sign` (for better cross-platform
+  compatibility) and optionally takes several of the same sub-parameters as
+  electron-osx-sign (#286)
 
 ### Deprecated
 
--- a/cli.js
+++ b/cli.js
@@ -27,6 +27,11 @@ if (args.tmpdir === 'false') {
   args.tmpdir = false
 }
 
+// (in this case, `Object` or `true`)
+if (args['osx-sign'] === 'true') {
+  args['osx-sign'] = true
+}
+
 packager(args, function done (err, appPaths) {
   if (err) {
     if (err.message) console.error(err.message)
--- a/mac.js
+++ b/mac.js
@@ -189,21 +189,17 @@ module.exports = {
           signOpts.platform = opts.platform
           signOpts.app = finalAppPath
 
-          // Take argument sign as signing identity:
-          // Provided in command line --sign, opts.sign will be recognized
-          // as boolean value true. Then fallback to null for auto discovery,
-          // otherwise provided signing certificate.
+          // Take argument osx-sign as signing identity:
+          // if opts['osx-sign'] is true (bool), fallback to identity=null for
+          // autodiscovery. Otherwise, provide signing certificate info.
           if (signOpts.identity === true) {
             signOpts.identity = null
           }
 
-          // console.log('signOpts', signOpts, signOpts.__proto__, process.cwd())
-
           sign(signOpts, function (err) {
             if (err) {
+              // Although not signed successfully, the application is packed.
               console.warn('Code sign failed; please retry manually.', err)
-            // Though not signed successfully, the application is packed.
-            // It might have to be signed for another time manually.
             }
             cb()
           })
--- a/mac.js
+++ b/mac.js
@@ -33,6 +33,10 @@ function filterCFBundleIdentifier (identifier) {
   return identifier.replace(/ /g, '-').replace(/[^a-zA-Z0-9.-]/g, '')
 }
 
+function signOptsWarning (name) {
+  console.warn(`WARNING: osx-sign.${name} will be inferred from main options`)
+}
+
 module.exports = {
   createApp: function createApp (opts, templatePath, callback) {
     var appRelativePath = path.join('Electron.app', 'Contents', 'Resources', 'app')
@@ -160,23 +164,46 @@ module.exports = {
         mv(sourceName, destName, cb)
       })
 
-      if (opts.sign) {
+      if ((opts.platform === 'all' || opts.platform === 'mas') && opts['osx-sign'] === undefined) {
+        console.warn('WARNING: signing is required for mas builds. Provide the osx-sign option, or manually sign the app later.')
+      }
+
+      if (opts['osx-sign']) {
+        // use default sign opts if osx-sign is true, otherwise clone osx-sign object
+        var signOpts = opts['osx-sign'] === true ? {identity: null} : Object.assign({}, opts['osx-sign'])
+
         operations.push(function (cb) {
-          sign({
-            app: finalAppPath,
-            platform: opts.platform,
-            // Take argument sign as signing identity:
-            // Provided in command line --sign, opts.sign will be recognized
-            // as boolean value true. Then fallback to null for auto discovery,
-            // otherwise provided signing certificate.
-            identity: opts.sign === true ? null : opts.sign,
-            entitlements: opts['sign-entitlements'],
-            'entitlements-inherit': opts['sign-entitlements-inherit']
-          }, function (err) {
+          // user may think they can pass platform or app, but they will be ignored
+          if (signOpts.platform) {
+            signOptsWarning('platform')
+          }
+          if (signOpts.app) {
+            signOptsWarning('app')
+          }
+          if (signOpts.binaries) {
+            console.warn('WARNING: osx-sign.binaries signing will fail. Sign manually, or with electron-osx-sign.')
+          }
+
+          // osx-sign options are handed off to sign module, but
+          // with a few additions from main options
+          signOpts.platform = opts.platform
+          signOpts.app = finalAppPath
+
+          // Take argument sign as signing identity:
+          // Provided in command line --sign, opts.sign will be recognized
+          // as boolean value true. Then fallback to null for auto discovery,
+          // otherwise provided signing certificate.
+          if (signOpts.identity === true) {
+            signOpts.identity = null
+          }
+
+          // console.log('signOpts', signOpts, signOpts.__proto__, process.cwd())
+
+          sign(signOpts, function (err) {
             if (err) {
               console.warn('Code sign failed; please retry manually.', err)
-              // Though not signed successfully, the application is packed.
-              // It might have to be signed for another time manually.
+            // Though not signed successfully, the application is packed.
+            // It might have to be signed for another time manually.
             }
             cb()
           })
--- a/mac.js
+++ b/mac.js
@@ -34,7 +34,7 @@ function filterCFBundleIdentifier (identifier) {
 }
 
 function signOptsWarning (name) {
-  console.warn(`WARNING: osx-sign.${name} will be inferred from main options`)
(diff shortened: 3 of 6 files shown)


malept:
For the moment, I want to get a consensus on what the API/CLI params should look like, so I did the docs first.

@positlabs per https://github.com/electron-userland/electron-packager/issues/285#issuecomment-194389616, I don't know what `bundle-identifier` is. That doesn't seem like the same as `identity`?

sethlu:
@malept Do we then feed the `osx-sign` object into `electron-osx-sign`? That could work quite nicely when put together.
Also, we could try the same mechanism for modifying the various plists' items in OS X builds?

malept:
> Do we then feed the `osx-sign` object into `electron-osx-sign`?

That was the plan, yes.

> Also, we could try the same mechanism for modifying the various plists' items in OS X builds?

That sounds good too, but that should wait until *after* 6.0.0 is released.

positlabs:
@malept, sorry, I meant `identity`, not `bundle-identifier`.

I kind of started working on this. Here is my progress, in case it helps: https://github.com/positlabs/electron-packager/blob/master/mac.js#L162-L166

There is some extra logic for handling `identifier`. I think ideally we would copy the whole `osx-sign` object and pass it to `electron-osx-sign` without modification. Maybe this warrants an update to `osx-electron-sign`?

sethlu:
On https://github.com/positlabs/electron-packager/blob/master/mac.js#L166, I think we may go with
```js
identity: signOpts.identity || true,
```
so when signing from CLI, we could simply use `--osx-sign` to trigger auto discovery of settings? Now we may need `--osx-sign.identity` instead.

positlabs:
Do you think this would work? I want to use the whole `osx-sign` object so it can serve as a passthrough to `electron-osx-sign`

```
      var signOpts = opts['osx-sign']

      // --osx-sign from CLI or osx-sign: true from API 
      if(signOpts === true){
        signOpts = {
          identity: true
        }
      }

      // assign additional options from main opts
      if(typeof signOpts === Object){
        signOpts.app = finalAppPath
        signOpts.platform = opts.platform
      }
```

malept:
I'm not a fan of having parameters be more than one type, but we've already got a precedent with `tmpdir`. As with `tmpdir`, there would need to be special handling of the argument in `cli.js` (because `minimist` can't handle an argument that's more than one type).

malept:
I updated the description of `osx-sign` in both the CLI and API docs. Let me know if those changes don't accurately reflect the discussion.

positlabs:
Looks good. Should we add docs for `binaries` and `verbose`?

malept:
...yeah, why not. It's not too much effort. I really want to feature freeze master so 6.0.0 can get out the door.

malept:
@positlabs do you mind writing the docs here (or in a branch of this branch)? I won't be able to get to it until tomorrow at the earliest.

positlabs:
Sure, I can do that now

positlabs:
I can't figure out how to grab this branch. Maybe I have to be a contributor? Anyway, here's a snippet from the `electron-osx-sign` docs.

```
`binaries` - *Array*

  Path to additional binaries that will be signed along with built-ins of Electron. Default to null.

`verbose` - *Boolean*

  Verbose flag, to display logs. Allowed values: true, false.
```

malept:
This is how I'd do it:
```
git clone https://github.com/electron-userland/electron-packager
git checkout osx-sign-single-param
```

malept:
I would prefer it if we have an abbreviated version electron-osx-sign docs (don't say what the defaults are, for example) and make it obvious that for more information, you need to look at the other module's docs. I think I already did that in the readme.

positlabs:
Huh, deleting the repo and re-cloning did the trick.

For the sake of brevity, and the fact that these options are documented elsewhere, I think we should only add docs for `binaries`.

sethlu:
I'm thinking if we need to here distinguish `darwin` and `mas` as the signing identities are different for the two if both specified from `osx-sign` (and I'll have a think on how to put it more naturally if to have these two settings separated).
Also, once we have the params consolidated, I could have these pages on https://github.com/electron-userland/electron-osx-sign/wiki updated.

sethlu:
Here's my interpretation of this current discussion: (Please correct me if any's misunderstood.)
- If to codesign an app with `electron-packager`, then a flag `--osx-sign` should be expected. (However, we are not taking into account that a `mas` distribution needs signing before going submitted.)
- If to codesign an app with `electron-packager`, however, with custom `identity`, then (without an explicit `--osx-sign` flag) `--osx-sign.identity` will call codesign with the identity plugged in.
On the `verbose` flag, I think it could go with `--verbose`, so `opts.verbose`; so with a `verbose` flag given for `electron-packager` (if it eventually is going to have something like it), the same should apply for `electron-osx-sign`.

malept:
> However, we are not taking into account that a mas distribution needs signing before going submitted.

We should address this before this gets merged.

Re: the `verbose` flag: I think your DEBUG change will be sufficient, I don't think we need to pass that in explicitly.

positlabs:
I have [added a few lines here](https://github.com/positlabs/electron-packager/tree/osx-sign-single-param), but I'm unsure of how to merge into this PR.

>(However, we are not taking into account that a mas distribution needs signing before going submitted.)

@sethlu I think we can allow `mas` builds to be unsigned in case the user wants to implement their own solution

malept:
@positlabs I'll merge it tomorrow. Looks good so far. Do you want to put your initial implementation in there as well?

sethlu:
> Re: the verbose flag: I think your DEBUG change will be sufficient, I don't think we need to pass that in explicitly.

@malept nice. I should be able to have https://github.com/electron-userland/electron-osx-sign/pull/24 merged so DEBUG could be supported in a later release.

> @sethlu I think we can allow mas builds to be unsigned in case the user wants to implement their own solution

@positlabs I don' know, because the entitlements specified should be different as well. :confused:

positlabs:
@sethlu is there a reason to sign `darwin` with entitlements? From what I understand, entitlements are only needed for `mas` builds, right? 

@malept I can start working on implementation. Likely push tomorrow morning

sethlu:
@positlabs It's not required to sign `darwin` with entitlements. However, as the same script is used for packaging `mas` and `darwin` builds, `--osx-sign` may be acknowledged for `mas` and `darwin` if to package for all platforms at a same time.

positlabs:
Oh, I hadn't considered that! 

I have some changes [here](https://github.com/positlabs/electron-packager/commit/4082edb544942cada0deac63b35b4fb7a25b187e). I tested building my app, and code signing is successful according to `spctl`.

Would there be any reason to allow the user to override `platform` or `app` if those options are included in `osx-sign`? Should we throw a warning/error if they attempt to? Or should we add logic to allow it? I vote we warn. If they want to really fiddle with the options, they can use `electron-osx-sign` directly.

malept:
> Would there be any reason to allow the user to override `platform` or `app` if those options are included in `osx-sign`? Should we throw a warning/error if they attempt to? Or should we add logic to allow it? I vote we warn. If they want to really fiddle with the options, they can use `electron-osx-sign` directly.

A warning would be sufficient.

----

So I think what I want to do now is let @positlabs work on their branch (based off of mine), and when they're done, I'll merge that into this branch in a separate PR, then rebase + merge the whole thing into master.

Yep, I know it's pretty convoluted.

positlabs:
@malept, thanks. I'll add that warning. 

Back to @sethlu's comment on mas distributions...

>(However, we are not taking into account that a mas distribution needs signing before going submitted.)

Should this be handled in a separate PR? I'm not sure how to address it.

malept:
I think it's just another warning.

positlabs:
I made a noob mistake and assumed `npm install` would pull in my local version of `electron-packager` even though version hadn't changed. My previous test was invalid.

`osx-sign` is borking on binaries because they are not done copying before it runs. I'm copying a binaries folder with the `asar-unpack-dir` option. Is there a way we can run this copy op before signing?

I added some warnings, and my code is here https://github.com/positlabs/electron-packager

malept:
I assume you mean https://github.com/positlabs/electron-packager/tree/osx-sign-single-param

You may want to run `$(npm bin)/standard`.

> Is there a way we can run this copy op before signing?

I don't see why not, from a conceptual standpoint. Try it and see if any tests fail.

positlabs:
@malet, yep, on that branch.

I've realized that the issue is that the files are actually in the temp build directory, and not the final path. 

One solution would be to resolve the `binaries` paths to the temp dir. This adds more logic, but I'm not sure of any other way.

@sethlu do you have any ideas?

sethlu:
@positlabs On the `asar-unpack-dir` option, I'd just recommend not to use the `asar` option so all unpacked binaries that come with one app could be completely signed individually and ready for shipping.
However, as you mentioned that the binaries are packed in `asar` packages. I can't really tell if those packed executables are permitted to run in a sandbox environment. (I'm not sure if `codesign` alters the internal structure of executables/binaries, or just stores a hash for testing data integrity.)

> Extra Unpacking on Some APIs
> Most fs APIs can read a file or get a file’s information from asar archives without unpacking, but for some APIs that rely on passing the real file path to underlying system calls, Electron will extract the needed file into a temporary file and pass the path of the temporary file to the APIs to make them work. This adds a little overhead for those APIs.
> From: http://electron.atom.io/docs/v0.36.3/tutorial/application-packaging/#extra-unpacking-on-some-apis

If those executables are unpacked to a temp dir, those binaries unsigned, it could be possible that they work without sandbox. I'll have a check on this now.

positlabs:
I swear I remember needing to sign these binaries for some reason, but maybe not. Using unsigned binaries in `asar-unpacked` doesn't seem to affect the `.pkg` or the `.app` during runtime, but I will try submitting a test build.

I'm using a lib, [node-fluent-ffmpeg](https://github.com/fluent-ffmpeg/node-fluent-ffmpeg), which uses `child_process.spawn()`, which requires the binaries to live in the `asar-unpacked` dir [according to this](http://electron.atom.io/docs/v0.36.0/tutorial/application-packaging/#executing-binaries-inside-asar-archive). 

I suppose I could go without asar, but I think this is an issue worth solving. I just submitted a test build with unsigned asar-unpacked binaries, and no errors came up in the upload process. I think that means the code signing was successful.

Is there ever a reason to sign extra binaries?

malept:
@positlabs is the `binaries` signing option the only thing holding up your branch? If so, I'd like to defer that to another PR.

positlabs:
@malept, yeah that's all. It just doesn't work because of temp dir path. I'm not even sure if it's really needed, but @sethlu might know more.

I can add a warning against it, and comment it out of the docs. I agree it shouldn't hold up v6 any more.

malept:
> I can add a warning against it, and comment it out of the docs. I agree it shouldn't hold up v6 any more.

:+1: Thanks for doing that. I'm going to (probably tomorrow) merge your branch into mine, squash your commits, and do some documentation fixes before merging this PR into master, and *finally* release v6.0.0.

malept:
In addition to documentation tweaks, I had to revert some syntax/API that didn't exist in Node 0.12, and added some testcases.

sethlu:
Just found https://github.com/electron/electron/blob/master/docs/tutorial/application-packaging.md#adding-unpacked-files-in-asar-archive. @positlabs I'll have a check later to see if some binaries present in the asar.unpacked dir will be auto-signed.

PR #289, opened 2016-03-12
Add examples of how to build for windows and linux in the readme

Glench:
Sorry for the drive-by pull request, but I think having very easy copy-pasteable lines for building for windows and linux would be *really* helpful for people using this system for the first time (like myself).

I may have gotten the commands wrong, in which case I would be happy to correct them if someone could tell me the right things.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -109,6 +109,18 @@ foobar
 
 The `FooBar.app` folder generated can be executed by a system running OS X, which will start the packaged Electron app.
 
+To generate the same app but for Windows, you could run:
+
+```
+electron-packager . FooBar --platform=win32 --arch=x64 --version=0.35.6
+```
+
+And for linux:
+
+```
+electron-packager . FooBar --platform=linux --arch=x64 --version=0.35.6
+```
+
 ### Programmatic API
 ```javascript
 var packager = require('electron-packager')


malept:
Thanks for the pull request! My thought is that instead of more examples, that we provide a tutorial, linked from the readme (because the readme is far too long as it is), which builds off of the [Electron tutorial](http://electron.atom.io/docs/v0.36.11/tutorial/quick-start/) - that would be a better fit for your changes. Let me know what you think of this proposal.

Glench:
Well, from my perspective as a new user coming to this project for the first time, these 3 lines are literally the only lines that matter for me (besides how to install electron-packager in the first place), so I'd tend toward putting them everywhere...in the readme, in the tutorial, wherever. I would love to be able to just copy and paste them and package my project for whatever platform immediately.

(I notice that I have a tendency in external libraries I use to skip all documentation and immediately look for an example of how to do the thing I'm trying to do. I've observed that lots of other people seem to have this tendency, too, and I'm wondering if you do.)

malept:
Adding more example lines to the readme just to illustrate that there is more than value for a parameter seems a little excessive to me. Why not add two more examples for the different arches for Windows/Linux as well?

Glench:
That could be done! I was just trying to show (concretely!) the most common use case so that people can basically copy and paste. I assumed the most common use case is "I want to compile this for Windows or Mac or Linux. How do I do that? Oh, here is an example." And two lines are pretty cheap in terms of document-length!

This particular kind of ease of use comes out my own experience using this library in particular and wanting examples, and from my experience using other libraries as well. For more on this (not from me) you might check out [09-19-15 | Library Writing Realizations](http://www.cbloom.com/rants.html).

malept:
if you want to deal with the common use case, the most obvious one is to just have one example showing `--all`. I imagine more people want to build for all platforms/arches rather than a specific one.

Glench:
Ooh, I didn't know there even was an --all option! That might be good! I like having a platform-specific one for my own use, but that might be good! Are you really worried about space? It's really not that much...

malept:
Yes, I am worried about space.

> I didn't know there even was an --all option!

That you didn't know about `--all` (it's documented in the readme, probably about halfway down) illustrates my assertion that the readme is too long. See also: #275.

Glench:
I think your conclusion is a little off. I didn't know there was an --all option because I didn't read the readme and rarely do. I only read until I find something that is immediately useful to what I'm trying to do. My task is to make a package, and the less I have to know about this library the better!

kfranqueiro:
> I imagine more people want to build for all platforms/arches rather than a specific one.

At least in my own personal use-cases, this is likely to become *less* of the case in 6.0.0, since I have no desire to tangle with MAS shenanigans (and is mainly why I wondered if adding that warranted the major bump).

> I didn't know there was an --all option because I didn't read the readme and rarely do.

I don't mean to sound snarky, but it seems a little funny to be suggesting a change to the readme if you're not going to actually read the readme...

Glench:
> I don't mean to sound snarky, but it seems a little funny to be suggesting a change to the readme if you're not going to actually read the readme...

On the contrary, it seems very natural. The documentation was missing something I needed, so I added it. It's only two lines, and I would argue they're the most useful lines in the entire argument.

malept:
I decided to go a different way with the readme. Thanks for reminding me that I needed to update the CLI example.

PR #294, opened 2016-03-30
Rename NEWS.md to CHANGELOG.md

lukeapage:
I completely missed news.md because I was looking for changelog and I didn't see it. (I actually found it because I was reading the commit history and found a commit called add news...)

Renaming will make it easier to find change changelog is alot more common.

wdyt?


malept:
I don't like calling it "changelog" because it has different meanings to different people. I'd rather keep it as "news" or call it "release notes".

lukeapage:
Ok fine, its just unusual. You might consider linking from the readme so its findable.

malept:
It's not that unusual. Compared to other projects, which use [History.txt](https://github.com/puma/puma/blob/master/History.txt) and [RELEASES.md](https://github.com/rust-lang/rust/blob/master/RELEASES.md), among other filenames, NEWS is one of the more common naming schemes.

>  You might consider linking from the readme so its findable.

Yep, that's the plan.

PR #296, opened 2016-03-31
Move API docs to its own file

### LINKED ISSUE #275: Reorganize docs

malept:
I cleaned up the headers, added some intra-doc links, and changed some formatting. I think it looks better, overall. And, you can now link to a specific option.

Fixes #275.

(Clicking "View" is probably a better way of looking at it, rather than the diff.)

## TODO

* [x] Update callback function signature (in two places)


docs/api.md
```diff
@@ -0,0 +1,233 @@
+# electron-packager API
+
+Short example:
+
+```javascript
+var packager = require('electron-packager')
+packager(options, function done_callback (err, appPath) { /* … */ })
```
kfranqueiro: The callback receives an array, since 5.0.0.  (I guess I missed updating something in the readme back when I did #88?)

malept: What does the callback function signature look like?

docs/api.md
```diff
@@ -0,0 +1,233 @@
...
+
+*Error* (or *Array*, in the case of an `ncp` error)
+
+Contains errors, if any.
+
+### `appPath`
+
+*String*
+
+Path to the newly created application.
```
kfranqueiro: This should be updated to an array of strings, and Paths to the newly created applications.  (Same as my note above, commenting here as well so we don't miss it)

docs/api.md
kfranqueiro: It's still error-object-first, it's just `(err, appPaths)` now rather than singular `appPath`.

docs/api.md
malept: Fixed in https://github.com/electron-userland/electron-packager/pull/296/commits/68413f8e3948cef759321510aef5f9eeccad6ea7.

PR #298, opened 2016-04-02
Delete default_app.asar during packaging

kevinsawicki:
In the next Electron release (0.37.4), the `default_app/` folder will now be a `default_app.asar` file.

This pull request adds support for deleting `default_app.asar` during packaging.

https://github.com/atom/electron/pull/4956

### DIFF
--- a/common.js
+++ b/common.js
@@ -111,7 +111,11 @@ module.exports = {
         fs.copy(opts.dir, appPath, {filter: userIgnoreFilter(opts), dereference: true}, cb)
       },
       function (cb) {
+        // Support removing old default_app folder that is now an asar archive
         rimraf(path.join(resourcesPath, 'default_app'), cb)
+      },
+      function (cb) {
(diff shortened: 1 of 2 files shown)


kevinsawicki:
Electron 0.37.4 is now out so `default_app.asar` is now present in that release.

malept:
We've just added a 0.37.4 fixture via #301. Would you mind adding a testcase using that test project to make sure that `default_app.asar` is deleted from an Electron release where it initially exists?

Once that's done, I'd like to merge this PR and release 6.0.1.

kevinsawicki:
@malept What Electron version do the basic tests run against by default? Is the current test in this pull request not sufficient?

malept:
@kevinsawicki the basic tests run against 0.35.4 or so, I believe due to the Mac App Store target platform.

malept:
I would rather have a test for Electron versions both before and after 0.37.4, to ensure both cases are covered.

PR #299, opened 2016-04-02
Add arm architecture

carapas:
Since there is a linux-arm build @ https://github.com/atom/electron/releases, it would make sense to allow arm architecture.


carapas:
Did not see there was a pending PR.

PR #301, opened 2016-04-03
Fix for issue 300: --icon option does not work under Electron 0.37.4

erkyrath:
https://github.com/electron-userland/electron-packager/issues/300

This is an easy fix. It writes the icon to the filename defined in CFBundleIconFile, so it will work in Electron 0.37.4 as well as earlier versions.

Since the test/config.json specifies "0.35.6", there's no way to test this fix directly. Is that okay, or should we figure out how to test both configurations? (If so, how?)

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -6,14 +6,21 @@ var series = require('run-series')
 var config = require('./config.json')
 var util = require('./util')
 
-// Download all Electron distributions before running tests to avoid timing out due to network speed
+// Download all Electron distributions before running tests to avoid timing out due to network speed.
+// Most tests run with the config.json version, but we have a special test (in mac.js) for 0.37.4.
 series([
   function (cb) {
     console.log('Calling electron-download before running tests...')
     util.downloadAll(config.version, cb)
   }, function (cb) {
+    console.log('Calling electron-download 0.37.4 before running tests...')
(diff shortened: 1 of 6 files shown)


malept:
I suggest creating a new (separate) test fixture project just for test/mac.js that has the Electron version set to 0.37.4, then add a testcase for the new `.icns` behavior.

erkyrath:
Okay, added fixture and test as described.

test/index.js
```diff
@@ -6,14 +6,21 @@ var series = require('run-series')
...
-// Download all Electron distributions before running tests to avoid timing out due to network speed
+// Download all Electron distributions before running tests to avoid timing out due to network speed.
+// Most tests run with the config.json version, but we have a special test (in mac.js) for 0.37.4.
 series([
   function (cb) {
     console.log('Calling electron-download before running tests...')
     util.downloadAll(config.version, cb)
   }, function (cb) {
+    console.log('Calling electron-download 0.37.4 before running tests...')
+    util.downloadAll('0.37.4', cb)
```
malept: It doesn't need to download Electron zips for 0.37.4, just the OSX ones.

malept: Never mind, there's more special changes in this release (see #298).

feross:
LGTM

PR #304, opened 2016-04-05
Added missed required option to readme

industral:


sbruchmann:
AFAIK the `version` parameter is optional as long as you’ve added `electron-prebuilt` to your `devDependencies`, because it will use its version.

Checking the [README](https://github.com/electron-userland/electron-packager/blob/91ef0f6e5e8268e6a524339016fe7393e18c41f1/readme.md) confirms my thoughts:

> - download all supported target platforms and arches of Electron using the installed electron-prebuilt version (and cache the downloads in ~/.electron)

Maybe this behaviour should be better documented.

industral:
Ok got it. Looks like it doesn't work when dependency has empty version

malept:
The `version` behavior is described in a few locations:
* the command line usage
* the example
* the API docs

That being said, I would happily review a PR that clarifies the docs. I don't see a need for this particular PR.

industral:
OK. But bug exists when user didn't specify version in package.json, e.g.

`"electron-prebuilt": "",`

then empty string returns false and code stop working. Checking key in object will fix it. See last commit.

malept:
That may be valid, but setting the version to a blank string is a really bad idea.

Please rename the PR and write a testcase for this change, and I will reopen this issue.

PR #311, opened 2016-04-08
Delete default_app.asar in Electron >= 0.37.4

malept:
This is #298 but with an extra unit test to make sure it works with Electron 0.37.4.


kevinsawicki:
Thanks for adding the spec :+1:

PR #315, opened 2016-04-09
Update rcedit to 0.5

develar:
fix https://github.com/atom/rcedit/pull/11


malept:
Looks like this means that we don't have to force users to install `MSVCP120.dll` anymore. (Even though that wasn't quite clear in the linked PR, sigh.)

PR #316, opened 2016-04-09
add mirror option

runinspring:
‘electron-download’ can use the mirror url for download

### DIFF
--- a/index.js
+++ b/index.js
@@ -101,6 +101,7 @@ function createSeries (opts, archs, platforms) {
         arch: arch,
         version: opts.version,
         cache: opts.cache,
+        mirror: opts.mirror,
         strictSSL: opts['strict-ssl']
       })
     })


malept:
Thanks for filing a pull request! Per our [contribution guidelines for pull requests](https://github.com/electron-userland/electron-packager/blob/master/CONTRIBUTING.md#filing-pull-requests), since you're adding a new option, it needs to be documented in both `docs/api.md` and `usage.txt`.

That being said, I'm beginning to think that we should have a `--download` flag that works like the `--osx-sign` or `--version-string`. `cache`, `mirror`, and `strictSSL` would be sub-options.

malept:
> That being said, I'm beginning to think that we should have a `--download` flag [...]

It bugged me enough that I wrote a PR for it this weekend: #320

Issue #150, opened 2015-10-06
Building packages on OS X silently fails during ncp call


davej:
Using electron-packager v5.1.0 on OS X 10.10.5 and Node v4.1.1

```
➜ electron-packager . MyApp --overwrite --platform=darwin --arch=x64 --version=0.33.6

Packaging app for platform darwin x64 using electron v0.33.6
```

No errors, it just exits about 10 seconds after displaying the message "Packaging app for platform darwin x64 using electron v0.33.6" and there are no files written to my CWD.

I realise that this is very sparse on info but I'm not sure what other details I can give. When I last ran this command about three weeks ago it worked fine.

malept:
Can you bisect which released version of electron-packager is the last one that works for you?

davej:
I believe I was using 5.1.0 when it worked before I tried moving back to an earlier version but it still didn't work. Could this be something related to something in my app code base (I see symlink issues are mentioned a lot, I have symlinked modules).

malept:
Could you try specifying an `--out` directory?

Otherwise, I think we're going to need reproduction steps on a minimal Electron app.

davej:
Still isn't working with `--out`.

I'm doing a bit of debugging, seems to be [failing here](https://github.com/maxogden/electron-packager/blob/d985624694b2132127ea88bae6f770790234824f/common.js#L82) the callback function (`cb`) is never fired.

malept:
Hmm. Very odd. I noticed that Travis CI only tests under Node 0.12. Let me add some more Node versions, and see if that surfaces anything.

davej:
Possibly related to https://github.com/AvianFlu/ncp/issues/98?

malept:
Possibly, but that's a really vague bug report with no follow-up.

davej:
Yup.

FYI, tried a couple of other node versions, I'm getting the same issue with Node-v0.12.2 and iojs-v2.0.1.

davej:
The actual copying of files works fine by the way, all the files are copied correctly, it's just that the callback isn't fired.

I put the callback in a `setTimeout` just to get it working and the app packaged itself correctly and runs as expected.

```javascript
  var ncpCallbackFired = false
  ncp(opts.dir, appPath, {filter: userIgnoreFilter(opts), dereference: true}, function(err) {
    ncpCallbackFired = true;
    cb(err);
  });
  // callback above isn't being fired by ncp for some reason,
[...]

PR #186, opened 2015-11-08
Ignore node_modules/.bin by default

malept:
I'm not sure there's a good reason that `node_modules/.bin` should be packaged along with the rest of the app. Besides, it breaks if you use NPM 3 and `--prune` (for why, see https://github.com/maxogden/electron-packager/issues/180#issuecomment-152897559).

This should fix #149 and #180.

Issue #151, opened 2015-10-07
swap out fs-extra.copy for tar-fs


maxogden:
to fix https://github.com/maxogden/electron-packager/issues/150

we should switch to using https://www.npmjs.com/package/tar-fs: `tarfs.pack().pipe(tarfs.extract())`, its like `JSON.parse(JSON.stringify(obj))` but for filesystems :) it's a simpler solution than using ncp, which is not currently maintained

patch welcome!

sindresorhus:
There's also [cpy](https://github.com/sindresorhus/cpy).

maxogden:
I wrote this for something else just now, its 12 lines: https://github.com/maxogden/cptar

natew:
Wondering if this would fix an issue I'm having...

When I run packager I get errors across a few apps. Chrome quits itself most fo the time and won't re-open without crashing until after packager runs. And then VSCode complains of too many open file references.

Guessing this is due to large amounts of node_modules. Currently am able to build most of the app into webpack bundles which may be the way to go.

Just realized this is a couple years old...

malept:
@natew Probably not. `tar-fs` is the transfer mechanism, it doesn't change the way the JavaScript is stored on disk.

malept:
I've been testing this enhancement off and on over the past week. The amount of code that has to change isn't bad, however there is a consistent performance decrease that I've noticed. It adds about a minute to the testsuite when I run it locally.

malept:
For reference: any replacement of `fs-extra.copy` has to have, at minimum, support for optionally dereferencing symlinks and a user-configurable filter.

malept:
I'm closing this due to the performance drop I saw.

PR #318, opened 2016-04-10
use fs-extra copy instead of ncp

### LINKED ISSUE #213: Stuck building for 10mins +
I am using this alongside with ember-electron package.

I found my build is taking over 12mins where i give up and cancel it. Stuck here:

```
$ electron-packager . FooBar --platform=darwin --arch=x64 --version=0.35.0
Packaging app for platform darwin x64 using electron v0.35.0
```
Nothing happens after that.

develar:
Closes #213, #150

electron-builder tests passed as well.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -220,7 +220,7 @@ Object (also known as a "hash") of application metadata to embed into the execut
 
 ### `err`
 
-*Error* (or *Array*, in the case of an `ncp` error)
+*Error* (or *Array*, in the case of an `copy` error)
 
 Contains errors, if any.
 
--- a/package.json
+++ b/package.json
@@ -24,7 +24,6 @@
     "fs-extra": "^0.26.5",
     "get-package-info": "0.0.2",
     "minimist": "^1.1.1",
-    "ncp": "^2.0.0",
     "object-assign": "^4.0.1",
     "plist": "^1.1.0",
     "rcedit": "^0.5.0",
--- a/test/fixtures/basic/package.json
+++ b/test/fixtures/basic/package.json
@@ -4,6 +4,8 @@
   "dependencies": {
     "run-series": "^1.1.1"
   },
+  "//": "ncp used to test https://github.com/electron-userland/electron-packager/pull/186",
+  "///": "(a module (with zero dependencies) that creates a file in node_modules/.bin)",
   "devDependencies": {
     "ncp": "^2.0.0",
     "run-waterfall": "^1.1.1",
--- a/mac.js
+++ b/mac.js
@@ -163,7 +163,7 @@ module.exports = {
               // Ignore error if icon doesn't exist, in case it's only available for other OS
               cb(null)
             } else {
-              ncp(icon, path.join(contentsPath, 'Resources', appPlist.CFBundleIconFile), cb)
+              fs.copy(icon, path.join(contentsPath, 'Resources', appPlist.CFBundleIconFile), cb)
             }
           })
         })
@@ -175,7 +175,7 @@ module.exports = {
         if (!Array.isArray(extras)) extras = [extras]
         extras.forEach(function (val) {
           operations.push(function (cb) {
-            ncp(val, path.join(contentsPath, 'Resources', path.basename(val)), cb)
+            fs.copy(val, path.join(contentsPath, 'Resources', path.basename(val)), cb)
           })
         })
       }
--- a/mac.js
+++ b/mac.js
@@ -1,6 +1,5 @@
 var common = require('./common')
 var fs = require('fs-extra')
-var ncp = require('ncp').ncp
 var objectAssign = require('object-assign')
 var path = require('path')
 var plist = require('plist')
--- a/test/basic.js
+++ b/test/basic.js
@@ -1,7 +1,6 @@
 var common = require('../common')
 var config = require('./config.json')
 var fs = require('fs-extra')
-var ncp = require('ncp').ncp
 var packager = require('..')
 var path = require('path')
 var series = require('run-series')
--- a/test/basic.js
+++ b/test/basic.js
@@ -398,7 +398,7 @@ function createIgnoreOutDirTest (opts, distPath) {
 
     series([
       function (cb) {
-        ncp(path.join(__dirname, 'fixtures', 'basic'), appDir, {dereference: true, stopOnErr: true, filter: function (file) {
+        fs.copy(path.join(__dirname, 'fixtures', 'basic'), appDir, {dereference: true, stopOnErr: true, filter: function (file) {
           return path.basename(file) !== 'node_modules'
         }}, cb)
       },
@@ -440,7 +440,7 @@ function createIgnoreImplicitOutDirTest (opts) {
 
     series([
       function (cb) {
-        ncp(path.join(__dirname, 'fixtures', 'basic'), appDir, {dereference: true, stopOnErr: true, filter: function (file) {
+        fs.copy(path.join(__dirname, 'fixtures', 'basic'), appDir, {dereference: true, stopOnErr: true, filter: function (file) {
           return path.basename(file) !== 'node_modules'
         }}, cb)
       },


docs/api.md
```diff
@@ -220,7 +220,7 @@ Object (also known as a "hash") of application metadata to embed into the execut
 
 ### `err`
 
-*Error* (or *Array*, in the case of an `ncp` error)
+*Error* (or *Array*, in the case of an `copy` error)
```
malept: Does `copy` actually create an `Array` on erroring? Because that would be very unfortunate.

package.json
```diff
@@ -36,10 +35,10 @@
   },
   "devDependencies": {
     "coveralls": "^2.11.6",
-    "nyc": "^5.6.0",
+    "nyc": "^6.2.1",
     "rcinfo": "^0.1.3",
     "run-waterfall": "^1.1.1",
-    "standard": "^5.4.1",
+    "standard": "^6.0.8",
```
malept: Do you mind removing these dependency updates from this PR?

test/fixtures/basic/package.json
```diff
@@ -5,7 +5,6 @@
     "run-series": "^1.1.1"
   },
   "devDependencies": {
-    "ncp": "^2.0.0",
```
malept: I added this particular dependency to test #186. If you remove it, please replace it with a module (with zero dependencies) that creates a file in `node_modules/.bin`.

malept:
@develar I looked into how `fs-extra` implemented `copy`, because you didn't change any of the options passed. It turns out that it's [just a wrapper for `ncp`](https://github.com/jprichardson/node-fs-extra/blob/d863f5fa00db751ff5faa73ae8a15344da677d29/lib/copy/copy.js#L3). Could you explain how this change would fix all the issues you mention?

malept:
Hm. I just realized that it uses a forked copy of ncp. But just because it may have some fixes in it, doesn't necessarily mean that all of those issues are resolved. We also don't have any tests for #150 or #213. If you want to assert that those are resolved, please write testcases for them.

package.json
develar: Strictly speaking, it is an unrelated change, so, sorry, I will remove and address in a separate PR.

docs/api.md
malept: I answered my own question below.

develar: We can pass `stopOnError: true` to stop on first error.

test/fixtures/basic/package.json
develar: reverted and comment added.

package.json
develar: reverted.

test/fixtures/basic/package.json
```diff
@@ -4,6 +4,8 @@
   "dependencies": {
     "run-series": "^1.1.1"
   },
+  "//": "ncp used to test https://github.com/electron-userland/electron-packager/pull/186",
+  "///": "(a module (with zero dependencies) that creates a file in node_modules/.bin)",
```
malept: That is a really unfortunate way to annotate JSON files.

develar: Yes, but npm doesn't support other ways (e.g. typescript supports comments in the tsconfig.json) :(

malept:
By the way, it might be better to squash your commits when the PR has been approved. It makes it easier for reviewers to see what changed, especially now that GitHub has added features to the pull request UI.

develar:
There is a comment https://github.com/AvianFlu/ncp/issues/98#issuecomment-173273082 so, I think, it is definitely fixed in the fs-extra. I don't see any particular file sets in the #150 and #213 so, I am not sure how I can test it. May be we can trust @jprichardson and existing fs-extra [tests](https://github.com/jprichardson/node-fs-extra/tree/master/lib/copy/__tests__)?

develar:
> it might be better to squash your commits

Ok, I will not amend commits anymore.

malept:
I'm not so sure the ncp issue you mention is fixed in fs-extra. If it is, it wasn't obvious from the commit log. My reading of his comment was more along the lines of "here's a maintained version of ncp".

jprichardson:
> I'm not so sure the ncp issue you mention is fixed in fs-extra. If it is, it wasn't obvious from the commit log. My reading of his comment was more along the lines of "here's a maintained version of ncp".

To my knowledge, I've fixed most of the outstanding issues in my forked copy of `ncp`. If there's anything you're unsure of, let me know and I can take a second look.

malept:
@jprichardson the issue I'm unsure that it fixes is https://github.com/AvianFlu/ncp/issues/98. Another look through the history (GitHub does not make that easy due to the multiple renames) indicates that it may be the same as https://github.com/jprichardson/node-fs-extra/issues/98 (coincidentally enough)?

jprichardson:
https://github.com/AvianFlu/ncp/issues/98 leaves out a lot of details, so it's unclear of the problem. The commenter just states that the callback is never being called with no code to reproduce. Who knows why. I put a ton of work into fixing the issues in `ncp` (as seen in https://github.com/jprichardson/node-fs-extra/issues/98) and others.

Also, I can state that I use `electron-packager` for my startup (http://www.exodus.io - depends upon Electron), so I'm heavily invested in the Electron ecosystem (if it matters).

docs/api.md
malept: We should probably update all of the `copy` calls to use that (in a separate PR). Having `err` be potentially more than one type is problematic for users.

malept:
@jprichardson @develar OK, sounds good to me. At minimum, it gets rid of an extra dependency (I've already gotten rid of direct production dependencies on `mv`, `rimraf`, and `mkdirp` in favor of `fs-extra` since this PR was filed).

@develar once you rebase, LGTM.

malept:
@develar FYI your build failed, seems [you missed some `ncp` references](https://travis-ci.org/electron-userland/electron-packager/jobs/122167908#L1628-L1629).

jprichardson:
Oh, I wanted to add one additional point. If it matters, `fs-extra` will be dropping support for Node v0.10 within the next few months. I wouldn't think this would be a big deal, but if you're looking to continue with Node v0.10 support (I couldn't imagine why), you'll be stuck in time with the last version of `fs-extra` to support Node v0.10. I don't see how this situation is any worse than the unmaintained `ncp` though. I just wanted to make sure you had all info up front before you make your decision.

malept:
It's been a while since we've supported 0.10 officially. I think I had to remove support for it in Travis CI for some feature. At any rate, I filed a PR the other day that [definitely removes support for Node < 4.0](https://github.com/electron-userland/electron-packager/pull/319). So we're OK :smile:

develar:
Sorry, fixed.

maxogden:
Just wanted to point out this doesnt close https://github.com/electron-userland/electron-packager/issues/151, that issue is about using tar-fs, which IMO is a lot more dependable performance wise than fs-extra/ncp.

malept:
Removed #151 from the issue summary.

malept:
@develar squash and fix the commit message to omit #151 and I will merge.

jprichardson:
> which IMO is a lot more dependable performance wise than fs-extra/ncp.

Would love to know why so that I can improve `fs-extra`.

Overall though, `tar-fs`, `cpy`, `fs-extra`... puh-tay-toe/puh-tah-toe. I don't care much as long as we get a better `electron-packager` and I'm happy to assist.

develar:
> which IMO is a lot more dependable performance wise than fs-extra/ncp.

It is a bit illogical for me. How does tar work — read file and append content to resulting (total) file. And then we will read total file and split it to destination files.

So, why it will be faster? copy operation just transfer bytes from one file to another.

Or do you mean that copy is not reliable? Well... if nobody cannot write reliable copy implementation on node platform... But I don't think so — I have tried 3 different implementations and only fs-extra copy works for me (electron-builder tests — copying test fixtures). And recently windows-installer got rid of jetpack and switched to fs-extra.

So, #151 was mentioned not by mistake.

develar:
Apart of this, I am going to get rid of copy phase (copy app files) at all :) Because if `prune` is not specified, we can build asar archive directly, without intermediate copy.

malept:
That "squash and merge" feature in GitHub's UI is pretty great. Won't work all the time, but worked well in this instance.

PR #319, opened 2016-04-11
Drop Node < 4 support

malept:
Removes a polyfill module and lets us use template strings.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ sudo: required
 dist: trusty
 language: node_js
 node_js:
-- '0.12'
 - '4'
 - '5'
 cache:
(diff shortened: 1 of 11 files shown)


malept:
I'll merge this in a couple of days (after I obviously fix some merge conflicts and squash) if there are no objections.

kfranqueiro:
Does this mean the next tagged version would end up being 7.0.0, or are you thinking of allowing this to go into a minor release (presumably not patch)?

malept:
The next release will be [7.0.0](https://github.com/electron-userland/electron-packager/milestones/7.0.0).

kfranqueiro:
Oh hey look at that milestone over there.  In that case, sounds good to me.

PR #320, opened 2016-04-11
Consolidate electron-download options

malept:
Consolidates `cache` and `strict-ssl` into the new `download` parameter, which is more or less passed directly to `electron-download` (similar to `osx-sign` and `version-string`). The original two parameters are now deprecated. This also lets us easily add sub-properties such as `mirror` (as requested in #316).

Depends on #319<del>, so [this](https://github.com/electron-userland/electron-packager/compare/drop-node-lt-4-support...malept:consolidate-download-options) might be a better diff to look at until it's merged</del>.

Closes #316.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -1,6 +1,19 @@
 #!/usr/bin/env node
 var fs = require('fs')
-var args = require('minimist')(process.argv.slice(2), {boolean: ['prune', 'asar', 'all', 'overwrite', 'strict-ssl', 'download.strictSSL']})
+var args = require('minimist')(process.argv.slice(2), {
+  boolean: [
+    'prune',
+    'asar',
+    'all',
+    'overwrite',
+    'strict-ssl',
+    'download.strictSSL'
+  ],
+  default: {
+    'strict-ssl': true,
+    'download.strictSSL': true
+  }
+})
 var packager = require('./')
 var path = require('path')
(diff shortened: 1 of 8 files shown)


malept:
This is in a better spot to be reviewed now, using the normal UI.

cli.js
```diff
@@ -1,6 +1,19 @@
...
+  boolean: [
+    'prune',
+    'asar',
+    'all',
+    'overwrite',
+    'strict-ssl',
+    'download.strictSSL'
+  ],
+  default: {
+    'strict-ssl': true,
```
malept: This actually fixes a bug with `strict-ssl` :cry:

feross: This should probably use the `alias` feature of minimist, so the two values are always in sync. Otherwise, the other default might overwrite the user setting.

feross: Yeah, wow. Good catch. This means that users using electron-packager v6 from the command line will accept invalid ssl certificates.

This is bad enough that I would recommend running `npm deprecate` on all older versions so users know to upgrade.

malept: All versions or just 6.x?

feross: Unclear.

malept: > This should probably use the `alias` feature of minimist, so the two values are always in sync. Otherwise, the other default might overwrite the user setting.

Hmm. I'll have to look into that, thanks.

PR #340, opened 2016-04-23
update fs-extra to 0.28

develar:


malept:
Any particular reason why?

develar:
Because `0` in the version means that `0.27` and `0.28` versions are not meet current range. Since electron-builder uses `0.28`, user gets two copy of `fs-extra`.

malept:
I'm aware of how semantic versioning works. I'm pretty sure the last time I looked at `npm list` for electron-packager, there were several packages with multiple versions installed.

So the only reason is to have it be the same version as electron-builder?

develar:
> So the only reason is to have it be the same version as electron-builder?

Yes, no strong reasons. Just keep deps up to date to avoid issues like https://github.com/electron/node-mksnapshot/pull/7

malept:
Fair enough. In the future, putting that in the PR summary would be recommended. (I still need to figure out the best thing to put in the pull_request_template.md.)

PR #341, opened 2016-04-23
Remove spaces from CFBundleExecutable key

### LINKED ISSUE #339: CFBundleExecutable can't have spaces
It looks like https://github.com/electron-userland/electron-packager/commit/f14c7ddf1b56e86fa729ce0ed5d93b96f52b4559 broke the packaging of my app.

I'm using an appname with a space in it and it looks like Mac doesn't like if that is used as the CFBundleExecutable. It works fine if I remove the space form the name.

"electron-packager": "7.0.1",
"electron-prebuilt": "0.37.6",
darwin, x64

develar:
Fixes #339.

and Squirrel.Windows bug "Windows shortcut doesn't work when `productName` contains a space" (https://github.com/electron-userland/electron-builder/issues/339)

Related issue: https://github.com/electron-userland/electron-builder/issues/288

Yes, this commit breaks electron-builder, but it will be fixed today in a separate commit (maybe some changes will be required on electron-packager side).

### DIFF
--- a/common.js
+++ b/common.js
@@ -1,3 +1,5 @@
+'use strict'
+
 var asar = require('asar')
 var child = require('child_process')
 var fs = require('fs-extra')
@@ -5,6 +7,7 @@ var minimist = require('minimist')
 var os = require('os')
 var path = require('path')
 var series = require('run-series')
+const sanitize = require('sanitize-filename')
 
 var archs = ['ia32', 'x64']
 var platforms = ['darwin', 'linux', 'mas', 'win32']
@@ -254,5 +257,9 @@ module.exports = {
     fs.stat(filename, function (err) {
       cb(err, err ? null : filename)
     })
+  },
+
+  sanitizeExeFilename: function (name) {
+    return sanitize(name.replace(/\s+/g, '_'))
   }
 }
--- a/mac.js
+++ b/mac.js
@@ -1,3 +1,5 @@
+'use strict'
+
 var common = require('./common')
 var fs = require('fs-extra')
 var path = require('path')
@@ -98,7 +100,7 @@ module.exports = {
       appPlist.CFBundleName = opts.name
       helperPlist.CFBundleDisplayName = opts.name + ' Helper'
       helperPlist.CFBundleIdentifier = helperBundleIdentifier
-      appPlist.CFBundleExecutable = opts.name
+      appPlist.CFBundleExecutable = common.sanitizeExeFilename(opts.name)
       helperPlist.CFBundleName = opts.name
       helperPlist.CFBundleExecutable = opts.name + ' Helper'
       helperEHPlist.CFBundleDisplayName = opts.name + ' Helper EH'
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "electron-download": "^2.0.0",
     "electron-osx-sign": "^0.3.0",
     "extract-zip": "^1.0.3",
-    "fs-extra": "^0.26.5",
+    "fs-extra": "^0.28.0",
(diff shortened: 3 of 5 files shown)


malept:
This is blocked on #340.

common.js
```diff
@@ -1,10 +1,13 @@
+'use strict'
+
```
malept: Can we discuss doing `'use strict'` in another PR?

develar: Well, since electron-packager requires nodejs 4.0, we just can use it (i.e. nothing to discuss here), if you insist, I can remove it.

malept: It's more of a consistency and standards issue. You're only putting it in *some* of the files. In terms of standards, I found some interesting GitHub issues around it, but it's irrelevant to this PR.

develar: In my change I need to add new variables. I decided to use `const`. And I cannot use `const`/`let` without `use strict`.  It was a reason why I use `use strict` here.  

Should I revert it and use only `var`?

malept: Yeah, I'd rather get everything in one pass.

common.js
```diff
@@ -254,5 +257,9 @@ module.exports = {
     fs.stat(filename, function (err) {
       cb(err, err ? null : filename)
     })
+  },
+
+  sanitizeExeFilename: function (name) {
```
malept: `exe` sounds awfully Windows-centric. `sanitizeExecutableFilename`?

Could you also write some tests for this function? (That don't involve running `packager()`.)

malept:
I'm not sure about the Windows change. If it's only a problem with Squirrel.Windows shortcuts, can't electron-builder deal with that?

common.js
malept: `use strict` conversation moved to #342.

develar:
> can't electron-builder deal with that?

Yes, it can be fixed on electron-builder side. But I think it will be useful in the electron-packager. 

On the other hand — yes, we should do avoid unnecessary escaping. I will revert.

develar:
The fix is not really correct — app `LEGO Digital Designer` works for me. Executable is `LEGO Digital Designer`. Hmm... I will investigate.

develar:
OS X — it is OS bug.
Windows — fixed in Squirrel.Windows.

About strict — I will prepare PR to use ES6 in my other PRs.

malept:
@develar By the way, I found out last night (via node.green) that you don't need `'use strict'` to use `const`.

develar:
@malept In the recent V8 VM (part of nodejs 6?), yes, strict mode is not required to use `let/const`. But we should support nodejs 4/5.

develar:
@malept http://v8project.blogspot.de/2016/01/v8-release-49.html "Release 4.9 also makes block level constructs such as class and let available outside of strict mode"

malept:
@develar you should look at http://node.green/

develar:
@malept Not quite understand what do you mean.

I see "is block-scoped (strict mode)"

```
function test() {
  let foo = 22;
}
```

```
/Users/develar/Documents/electron-packager/foo.js:2
  let foo = 22;
  ^^^

SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:387:25)
    at Object.Module._extensions..js (module.js:422:10)
    at Module.load (module.js:357:32)
    at Function.Module._load (module.js:314:12)
    at Function.Module.runMain (module.js:447:10)
    at startup (node.js:148:18)
    at node.js:405:3
```

malept:
Oh, I read node, not v8.

PR #344, opened 2016-04-24
Escape dot in ".git" correctly

feross:
Previously, folders name “agit” or “bgit" would get ignored.


feross:
Caught this issue with a new rule I'm considering for `standard`: http://eslint.org/docs/rules/no-useless-escape

malept:
Ha, nice. :+1:

PR #351, opened 2016-04-27
'use strict' in all of the files #342

### LINKED ISSUE #342: 'use strict' in all of the files
Moving conversation from #341.

I am not against using `'use strict'`, but if we do so, I'd like to consider switching to using `strict-standard` instead of `standard`. Or possibly `eslint` + `eslint-config-standard` + some strict config (the reasoning being that it's more likely that a given editor will have eslint support available more readily than `standard` - this may just mean that we keep some form of `standard` for `npm test`, and just add an `.eslintrc` to the project).

See also:

* https://github.com/feross/standard/issues/382

develar:
I want to make some improvements in the electron-packager code. Since we require nodejs 4, we can use `let` and want use it.

To avoid unrelated changes in the PRs, I forced to make this PR :)
Fixes #342


.eslintrc
```diff
@@ -1,3 +1,11 @@
...
-  "extends": "standard"
+  "extends": "standard",
+  "parserOptions": {
+    "sourceType": "script"
+  },
+  "rules": {
+    "strict": [
+      "error"
+    ]
+  }
```
malept: I was reading last night that you can also put eslint config in `package.json` - perhaps we should do that.

.eslintrc
```diff
@@ -1,3 +1,11 @@
 {
-  "extends": "standard"
+  "extends": "standard",
+  "parserOptions": {
+    "sourceType": "script"
```
malept: I know this is necessary for the `strict` rule to work correctly, but I don't think it's accurate - most of the files are modules, right?

.eslintrc
develar: Yes, I also want it. But I see that you use some editor integration — "Editor integration is easier that way." If your editor support eslintrc in the package.json, we definitely should get rid of yet another file and move configuration to package.json

.eslintrc
develar: The issue — standard config is not correct. See https://github.com/feross/eslint-config-standard/blob/master/eslintrc.json#L8 

@feross Could you please comment why it is so? 

https://github.com/eslint/eslint/issues/5301#issuecomment-184720161 is quite clear. We should not set `module` — we don't use ES6.

feross: @develar `standard` always lints files as modules so that users don't need to specify which mode they're in. There's little downside to always running in module mode.

.eslintrc
malept: @develar per http://eslint.org/docs/user-guide/configuring#configuration-file-formats, eslint supports `package.json` - if some editor only supports `.eslintrc`, then it needs to be fixed.

develar:
@malept Friendly ping :)

malept:
@develar I took a small break from reviewing PRs. I'll try to get to it this week.

develar:
In my fork I use promises — see mac.js for example https://github.com/develar/electron-packager/blob/master/mac.js We don't use sync methods anymore, and code a little bit simplified. It is why I prepare this PR because otherwise I cannot create PR about promises ;)

develar:
In my fork I use promises — see mac.js for example https://github.com/develar/electron-packager/blob/master/mac.js We don't use sync methods anymore, and code a little bit simplified. It is why I prepare this PR because otherwise I cannot create PR about promises ;)

.eslintignore
```diff
@@ -0,0 +1,2 @@
+**/test/fixtures/**/node_modules/**
+**/test/work/*
```
malept: I didn't really want to add another dotfile but it seems that eslint has technical reasons to keep this one: https://github.com/eslint/eslint/issues/4261

*(The "cascading" reason is a bit silly IMO though. The execution one is quite reasonable.)*

malept:
Sorry for the delay, work and real life got in the way.

.eslintignore
develar: There is no technical reason to not support specifying ignore in the package.json, but eslint doesn't support it.

> The execution one is quite reasonable

For config file yes, but we specify config in the `package.json`

malept: > There is no technical reason to not support specifying ignore in the package.json, but eslint doesn't support it.

From [the comment right above the one you posted in the linked issue](https://github.com/eslint/eslint/issues/4261#issuecomment-213461874):
> There's also a problem with timing, we need ignore files before we start traversing and we only find config files while we are traversing.

That's what I meant by execution reason.

PR #353, opened 2016-04-29
Move asar options to sub-parameters passed through asar-options

robfuscate:
This will let us pass an ordering file to the asar process.

### DIFF
--- a/common.js
+++ b/common.js
@@ -214,6 +214,9 @@ module.exports = {
         if (opts['asar-unpack-dir']) {
           asarOptions.unpackDir = opts['asar-unpack-dir']
         }
+        if (opts['asar-ordering']) {
+          asarOptions.ordering = opts['asar-ordering']
+        }
         asarApp(path.join(appPath), asarOptions, cb)
       })
     }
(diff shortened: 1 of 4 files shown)


robfuscate:
The asar readme doesn't mention this option, but the changelist where it was added is here https://github.com/electron/asar/pull/52

malept:
> The asar readme doesn't mention this option

😞 Someone should fix that.

As a general comment to this PR, I am not opposed to this (given that there are docs added and possibly tests), but I'd rather have all of the asar options work like `version-string`, `osx-sign`, and now `download`, that is, grouped as sub-parameters. Obviously this would mean a major version bump, since there would be a breaking parameter change to `asar`.

robfuscate:
By moving things to asar-options, we've essentially made it so you can pass arbitrary flags to asar. I've renamed the pull request.

malept:
It would be nice if `asar-options` was merged into `asar` - that is, `asar` can be either `true` or an object of options.

Also,  you should keep the existing `asar-unpack`* documentation and mark it deprecated (as I've done for `strict-ssl`, etc.)

malept:
There should also be tests to make sure that `asar` options override the deprecated `asar-unpack`, etc. options.

malept:
Continued in #417.

Issue #270, opened 2016-02-21
Feature Request: non-patented FFMPEG Option


arcticShadow:
Hi, 
In the last few days, Electron has been released with ffmpeg [dynamically linked,](https://github.com/atom/electron/releases/tag/v0.36.8) with the ability to swap out FFMPEG with an instance that does not contain patented codecs. 

Would it be possible to provide a feature in this packager to swap out this library during the build process. A quick scan of the source shows that the electorn-downloader pulls in the electorn prebuilt. and extracts it, so it seems like it would be trivial to add a second download step to grab the ffmpeg binary  and replace the existing one. 

I'd be happy to look at a PR if this feature holds merit.

maxogden:
Seems like a good idea, but my concerns would be:

- how much code it would add here
- if we can add an api to let people define third party packaging steps
- can it be implemented in a third party that uses o/

arcticShadow:
I don't imagine that much code would be added here. Although, I do like the idea of extending the build process to 3rd party apps. This would potentially open up quite a big customisation opportunity. 

I'll have a think about how this can be implemented, and try do a code dive to figure out the where and how.

maxogden:
Cool sounds good!

arcticShadow:
Done some digging into how this could be achieved. 

I think the best route to enable 3rd party build steps will be: 
* Plugin loader system. All plugins defined as options will be loaded during bootstrap phase. 
* add a "PostBuild" method to common.js (or possibly, a generic method, that takes a build step as a param, and an object/array of options)
* In each linux/mac/win file, after the package/copy process, but before any signing or other final tasks, call the above methods from common, and pass in any relevant args. 
* 3rd party plugin makes changes, and returns to the platform specfic file to continue. 


This feels like it could be a good use of generators. Thoughts?

malept:
Brief searching indicates that generators have been in Node since ~0.11. Is this correct?

arcticShadow:
That sounds right. It would be behind the `--harmony` flag though. Sometimes I forget what is and is not ES5 safe code.

malept:
Hmm. I'm not sure if that's allowable, then.

kfranqueiro:
It's listed as being included with no flags required at this point (though it's unclear to me whether this list pertains to both 4.x and 5.x): https://nodejs.org/en/docs/es6/

malept:
[...]

PR #354, opened 2016-04-30
Implemented an 'afterExtract' function hook  RE: #270

MarshallOfSound:
This PR adds an `afterExtract` option to the API.  This option should be an array of functions, each of which is called directly after `electron-packager` has finished extracting to the temporary directory.

Relevant issue: #270

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -45,6 +45,18 @@ The non-`all` values correspond to the platform names used by [Electron releases
 
 #### All Platforms
 
+##### `afterExtract`
+
+*Array of Functions*
+
+An array of functions to be called after Electron has been extracted to a temporary directory.  Each function is called with five parameters:
+
+- `buildPath` (*String*): The path to the temporary folder where Electron has been extracted to
+- `electronVersion` (*String*): The version of electron you are packaging for
+- `platform` (*String*): The target platform you are packaging for
+- `arch` (*String*): The target architecture you are packaging for
+- `callback` (*Function*): Must be called once you have completed your actions
+
 ##### `all`
 
 *Boolean*
--- a/index.js
+++ b/index.js
@@ -128,6 +128,16 @@ function createSeries (opts, archs, platforms) {
             },
             function (cb) {
               extract(zipPath, {dir: buildDir}, cb)
+            },
+            function (cb) {
+              if (!opts.afterExtract || !Array.isArray(opts.afterExtract)) {
+                cb()
+              } else {
+                var newFunctions = opts.afterExtract.map(function (fn) {
+                  return fn.bind(this, buildDir, version, platform, arch)
+                })
+                series(newFunctions, cb)
+              }
             }
           ], function () {
             require(supportedPlatforms[platform]).createApp(comboOpts, buildDir, callback)
--- a/test/hooks.js
+++ b/test/hooks.js
@@ -1,0 +1,60 @@
+'use strict'
+
+var config = require('./config.json')
+var fs = require('fs')
+var packager = require('..')
+var path = require('path')
+var series = require('run-series')
+var test = require('tape')
+var util = require('./util')
+var waterfall = require('run-waterfall')
+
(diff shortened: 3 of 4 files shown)


index.js
```diff
@@ -125,7 +125,16 @@ function createSeries (opts, archs, platforms) {
...
+              extract(zipPath, {dir: buildDir}, function (err) {
+                if (err || !opts.afterExtract || !Array.isArray(opts.afterExtract)) {
+                  cb(err)
+                } else {
+                  var newFunctions = opts.afterExtract.map(function (fn) {
+                    return fn.bind(this, buildDir, version)
+                  })
+                  series(newFunctions, cb)
+                }
+              })
```
malept: I think the callback you defined for `extract` be another function in the series.

docs/api.md
```diff
@@ -45,6 +45,16 @@ The non-`all` values correspond to the platform names used by [Electron releases
 
 #### All Platforms
 
+##### `afterExtract`
+
+*Array*
```
malept: Should probably read

> *Array* of *Function*s

docs/api.md
```diff
@@ -45,6 +45,16 @@ The non-`all` values correspond to the platform names used by [Electron releases
...
+
+An array of functions to be called after Electron has been extracted to a temporary directory.  Each function is called with three parameters:
+
+* `buildPath` - The path to the temporary folder where Electron has been extracted to
+* `electronVersion` - The version of electron you are packaging for
+* `callback` - Must be called once you have completed your actions
```
malept: Minor nit: the rest of the file has standardized on `-` for list items.

index.js
MarshallOfSound: Are you suggesting then that we should do

```js
series([extract(zipPath, {dir: buildDir})].concat(newFunctions))
```

docs/api.md
```diff
@@ -45,6 +45,18 @@ The non-`all` values correspond to the platform names used by [Electron releases
 
 #### All Platforms
 
+##### `afterExtract`
+
+*Array*
+
+An array of functions to be called after Electron has been extracted to a temporary directory.  Each function is called with three parameters:
```
malept: > five parameters

*(Insert Spanish Inquisition meme here)*

MarshallOfSound: My counting skillz are amazing :+1:

index.js
malept: I was thinking more along the lines of

```javascript
function (cb) {
  extract(zipPath, {dir: buildDir}, cb)
},
function (cb) {
  if (opts.afterExtract exists and is not array) {
    return error
  }
  // etc.
}
```
...that got pseudocode-y, fast.

MarshallOfSound: Ah got it :+1: 

Done

malept:
Looks good, I think this just needs at least one unit test.

MarshallOfSound:
@malept Not too familiar with the testing framework here but I added some tests :+1:

malept:
The test logic looks OK at a glance (I'll look at it in more detail tomorrow), but I'd prefer if it had its own test instead of tacking it onto another test.

MarshallOfSound:
@malept I've moved the logic to it's own test and moved it to a new file called "hooks" so that if in the future we add more of these hooks we have a place for them to go :+1:

malept:
Sorry about the delay, work & real life got in the way.

I had merged #351 before this one, so if you could take care of the `strict` changes in `test/hooks.js`, I'll merge.

MarshallOfSound:
@malept Talk about delay 😆 

I finally got around to adding `use strict` 😁

malept:
Amusingly enough, this morning I did something similar. The only difference is that the require statements are `const`:
https://github.com/electron-userland/electron-packager/tree/after-extract-hook

MarshallOfSound:
@malept So who's do we use 😆

malept:
I added a PR for my branch (#403). I'll wait for CI to run and then merge it.

PR #364, opened 2016-05-13
enable `asar` and `prune` options by default - fixes #35

### LINKED ISSUE #35: `--asar` and `--prune` should be enabled by default
I don't see why you wouldn't want those and I think most do. We could instead have `--no-asar` and `--no-prune` flags. I much prefer us to have sane defaults, which means less config mess for users.

sindresorhus:
// @malept @develar

### DIFF
--- a/common.js
+++ b/common.js
@@ -167,8 +167,8 @@ module.exports = {
     // * Creates temporary directory
     // * Copies template into temporary directory
     // * Copies user's app into temporary directory
-    // * Prunes non-production node_modules (if opts.prune is set)
-    // * Creates an asar (if opts.asar is set)
+    // * Prunes non-production node_modules
+    // * Creates an asar
 
     var tempPath
     if (opts.tmpdir === false) {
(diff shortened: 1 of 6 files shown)


develar:
👍 asar is already default in the `electron-builder`. And it should be default because Window IO is very poor and asar helps a lot on Windows platform.

But please note — https://github.com/electron-userland/electron-builder/issues/351 (anyway, it is electron-builder responsibility to deal with it).

malept:
A few general notes:

* This requires a major version bump.
* I don't think this sets either parameter by default for the CLI.
* This will conflict with #353. (And I want to take that one first because it should only require a minor version bump.)

malept:
Also looks like there's at least one failing test.

develar:
@sindresorhus In the next upcoming electron-builder 5 `prune` will be enabled by default if not two-package.json project structure is used.

sindresorhus:
Closing this. I'm now using `electron-builder` which has `asar: true` by default. And I don't have the time to finish this PR.

PR #374, opened 2016-05-20
Remove hard dependency on electron-prebuilt

mcfedr:
This means packager will also use electron version specified in `electron-version` field in package.json

This is good because we don't have to depend on `electron-prebuilt`. `prebuilt` automatically downloads
electron for the current platform, but `packager` might only be building for a different platform, and will use
`electron-download` anyway, so we can avoid unnecessary download, and speed up `npm install` time.

### DIFF
--- a/index.js
+++ b/index.js
@@ -41,7 +41,7 @@ function validateList (list, supported, name) {
 function getNameAndVersion (opts, dir, cb) {
   var props = []
   if (!opts.name) props.push(['productName', 'name'])
-  if (!opts.version) props.push(['dependencies.electron-prebuilt', 'devDependencies.electron-prebuilt', 'electron-version'])
+  if (!opts.version) props.push(['dependencies.electron-prebuilt', 'devDependencies.electron-prebuilt', 'config.electron-version'])
 
   // Name and version provided, no need to infer
   if (props.length === 0) return cb(null)
--- a/index.js
+++ b/index.js
@@ -41,7 +41,7 @@ function validateList (list, supported, name) {
 function getNameAndVersion (opts, dir, cb) {
   var props = []
   if (!opts.name) props.push(['productName', 'name'])
-  if (!opts.version) props.push(['dependencies.electron-prebuilt', 'devDependencies.electron-prebuilt'])
+  if (!opts.version) props.push(['dependencies.electron-prebuilt', 'devDependencies.electron-prebuilt', 'electron-version'])
 
   // Name and version provided, no need to infer
   if (props.length === 0) return cb(null)
--- a/index.js
+++ b/index.js
@@ -41,7 +41,7 @@ function validateList (list, supported, name) {
 function getNameAndVersion (opts, dir, cb) {
   var props = []
   if (!opts.name) props.push(['productName', 'name'])
-  if (!opts.version) props.push(['dependencies.electron-prebuilt', 'devDependencies.electron-prebuilt', 'config.electron-version'])
+  if (!opts.version) props.push(['dependencies.electron-prebuilt', 'devDependencies.electron-prebuilt'], 'config.electron-version')
 
   // Name and version provided, no need to infer
   if (props.length === 0) return cb(null)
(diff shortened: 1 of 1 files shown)


develar:
@mcfedr What do you use during development? Or you just not specify `electron-prebuilt` dependency but install it manually?

As far I understand, `--production` flag is not a solution because we need to install babel/compiler/etc.

I think, we should not reinvent the wheel and introduce new non-standard field in `package.json`. Instead, we should use [config](https://docs.npmjs.com/misc/config#per-package-config-settings).

```
"config": {
    "electron-version": "",
  },
```

mcfedr:
Config works for me, ill change it now.

I current specify `electron-prebuilt` because it doesn't work without it, but I want to avoid it downloading when I do `npm install`, for example, my tests are run and fail, I didnt get to the electron stage, and saved all that time downloading

develar:
>  I didnt get to the electron stage, and saved all that time downloading

I don't think that it worth. 

* electron don't want to distribute using normal compression — https://github.com/electron/electron/issues/5455 Well, because in the modern world 50 vs 25 mb means nothing.
* Travis supports cache on OS X now, so, just configure cache. https://github.com/electron-userland/electron-builder/blob/master/.travis.yml#L22 Or circleci — https://github.com/develar/electron-builder/blob/builder/circle.yml#L7

So, nothing to save. It is downloaded only once. It is not unreliable and slow npm, it is downloaded quickly :)

mcfedr:
The other example is I am wanting to deploy my web app version of the code, that completely doesn't require electron.

mcfedr:
Either way, this surely does no harm :)

Also right now, `electron-prebuilt` hasnt been updated to 1.1.1, put using this, it would be downloaded and everything would work...

malept:
I'm not opposed to adding this feature in principle. Per the [contributing document](https://github.com/electron-userland/electron-packager/blob/master/CONTRIBUTING.md), it needs tests and documentation. In fact, I think the change broke at least one existing test.

On the other hand, I'm a bit hesitant because I think the documentation for specifying the Electron version is going to get fairly complicated.

mzmousa:
@develar @malept I'm trying to use the config solution and have added
```
"config": {
    "electron-version": "1.2.1"
  },
```
in my `package.json`. However, I'm unsure as to where I would use this config option. Do I add something after the npm script (the one that's build -o or -w or -l)? Or do I pass it into ```build.osx``` as a variable of sorts?

develar:
@mzmousa If you use electron-builder please use `build.electronVersion` — https://github.com/electron-userland/electron-builder/pull/418#issuecomment-221798404 

If I understand correctly, @mcfedr created PR here by mistake, because electron-builder in any case use electron-builder in a way that prevent any such configuration.

malept:
If this PR is no longer wanted by the submitter, that would be good to know.

mcfedr:
@develar Its not a mistake here, i think the feature makes sense for both projects, builder and packager. 

And it would have been nice to have the same way of doing it for both, via `config.electron-version`, although now electron-builder has released `build.electronVersion`

Unfortunately I haven't had time to update this pull request with the needed tests and docs changes.

mzmousa:
Ah, I should have just looked at the contributing document(s). Works fine for me, thanks!

develar:
@mcfedr electron-builder forbids most of the electron-packager options to avoid misconfiguration. https://github.com/electron-userland/electron-packager/blob/master/docs/api.md#version is forbidden as well (yes — you will get build error). 

electron-packager currently doesn't support configuration via `package.json`. If you want to set electron version — you have to use CLI param `version` or programmatic API.

>  i think the feature makes sense for both projects, builder and packager.

My previous suggestion is wrong — `config` is not suitable here because there is an obvious question — why I can specify electron version using package.json, but other options only using CLI?
And so, another question is — should electron-packager introduce `build` as electron-builder does?

I have no answers. For me electron-packager is a low-level tool and electron-builder is a orchestration tool. And in any case, even if electron-packager maintainers decided to introduce `build`, electron-builder in any case will not support all options — but forbid it (e.g. arch).

malept:
I seriously doubt that electron-packager is going to support a top-level `build` key to `package.json` (@develar even linked to the reason why [above](#issuecomment-220567321)). However, I would not be opposed to a discussion about adding a sub-key to `config` (e.g., `config.electronPackager`) which basically looks like the parameters for `packager()`. It would be wise to discuss that option before this PR gets merged.

RedSparr0w:
you do not need `electron-prebuilt` as a dependency all you have to do is specify a version in the `--version` param

example:
`electron-packager "./" "My App" --platform=win32 --arch=x64 --version=1.2.5`

malept:
Since there are outstanding problems with this pull request, and the original submitter has stated that they don't have time to work on it, I'm going to close this for now.

PR #376, opened 2016-05-25
Added the ability to use .electronignore files to ignore on building

jasonfill:
For our particular use case we had quite a few files we wanted to ignore and the ability to add these to a file that could be tracked within our repo was a more comfortable option rather than modifying the actual build scripts each time we needed to add something we wanted to ignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ node_modules
 test/work
 .DS_Store
 .nyc_output
+.idea/*
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+## [7.0.3] - 2016-05-25
+
+### Added
+
+* Add support to use `.electronignore` and `.electronignore_{platform}` files.
 
 ## [7.0.2] - 2016-05-18
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "electron-packager",
-  "version": "7.0.2",
+  "version": "7.0.3",
   "description": "Package and distribute your Electron app with OS-specific bundles (.app, .exe etc) via JS or CLI",
   "main": "index.js",
   "bin": {
--- a/test/basic.js
+++ b/test/basic.js
@@ -248,6 +248,34 @@ function createIgnoreTest (opts, ignorePattern, ignoredFile) {
   }
 }
 
+function createIgnoreFromGlobalFileTest (opts, ignoredFile) {
+  return function (t) {
+    t.timeoutAfter(config.timeout)
+
+    opts.name = 'basicTest'
+    opts.dir = path.join(__dirname, 'fixtures', 'basic')
+
+    var appPath
+
+    waterfall([
+      function (cb) {
+        packager(opts, cb)
+      }, function (paths, cb) {
+        appPath = path.join(paths[0], util.generateResourcesPath(opts), 'app')
+        fs.stat(path.join(appPath, 'package.json'), cb)
+      }, function (stats, cb) {
+        t.true(stats.isFile(), 'The expected output directory should exist and contain files')
+        fs.exists(path.join(appPath, ignoredFile), function (exists) {
+          t.false(exists, 'Ignored file should not exist in output app directory')
+          cb()
+        })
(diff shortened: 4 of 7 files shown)


.gitignore
```diff
@@ -6,3 +6,4 @@ node_modules
 test/work
 .DS_Store
 .nyc_output
+.idea/*
```
malept: You should put that in your global .gitignore instead.

NEWS.md
```diff
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+## [7.0.3] - 2016-05-25
```
malept: Don't put a new version number here, just the "Added"  section is sufficient.

This would be a minor version bump, anyway.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "electron-packager",
-  "version": "7.0.2",
+  "version": "7.0.3",
```
malept: Don't bump the version number.

test/basic.js
```diff
@@ -549,6 +577,8 @@ util.testSinglePlatform('ignore test: string in array', createIgnoreTest, ['igno
 util.testSinglePlatform('ignore test: string', createIgnoreTest, 'ignorethis', 'ignorethis.txt')
 util.testSinglePlatform('ignore test: RegExp', createIgnoreTest, /ignorethis/, 'ignorethis.txt')
 util.testSinglePlatform('ignore test: Function', createIgnoreTest, function (file) { return file.match(/ignorethis/) }, 'ignorethis.txt')
+util.testSinglePlatform('ignore test: Function', createIgnoreFromGlobalFileTest, 'ignorethis.txt')
```
malept: Should be named differently from the other test.

malept:
What do you think about adding a key to `package.json` instead? I know I don't want to deal with the idea of more dotfiles in my repositories if I don't have to. See: https://docs.npmjs.com/files/package.json#config

test/basic.js
malept: Also you should use another filename for this test, otherwise I believe it will create a false positive for the other ignore tests.

jasonfill:
@malept All good suggestions, I will address and re-submit.  It does not look like the builds failed due to my code changes - although I could be wrong. Is it somewhat standard for these builds to fail? I would think not but wanted to ask.

malept:
>  Is it somewhat standard for these builds to fail?

Depends on the test. The tests that shell out to other programs (wine, OSX signing) have a tendency to time out and I don't know why. I need to fix into the OSX errors (I hadn't seen that before, looks like Travis updated their OSX image?), but the Linux failures look legit (see my comment about the test).

develar:
@malept `xcode7.3` is broken and slow (or often hangs). electron-builder switched back to `xcode7` (I haven't yet time to contact Travis support).

malept:
@develar good to know, I'll mess with different OSX images on a branch (when I have time) and see what the execution times are.

develar:
@malept FYI: https://blog.travis-ci.com/2016-04-27-xcode73-performance

test/basic.js
malept: Keep in mind when you refactor your code to also add a test for the platform-specific ignores as well.

malept:
@jasonfill I would suggest this:

```javascript
{
// ...
  "config": {
    "electron-packager": {
      "ignore": ["foo", "bar"]
    }
  }
// ...
}

craigcosmo:
Why don't we have both option? We can either config in the package.json and .packagerignore file, it will be more flexible for sure.

malept:
I don't want to have both options because of my reason in https://github.com/electron-userland/electron-packager/pull/376#issuecomment-221579694, and more importantly, because it's not obvious which option takes precedence if both methods are used, and will violate the principle of least astonishment either way.

jasonfill:
@malept We actually ended up using the API with the [ignore option](https://github.com/electron-userland/electron-packager/blob/master/docs/api.md#ignore) and it worked perfect and a bit cleaner than this anyway. So at this point I am going to close this pull request.

englishextra:
```.electronignore``` filename isnt obvoius whether its for running or building

PR #381, opened 2016-05-30
temporary fix for node 6 stdio issue

juliangruber:
Because of https://github.com/nodejs/node/issues/6456, in node@6 you don't (yet) see all of the usage text because the process exits before all was written out. This is a temporary fix.


malept:
Thanks!

PR #395, opened 2016-06-13
Document ignore using Array[RegExp]

davej:
I wasn't aware you could do this until I looked through the code.


malept:
Thanks for prompting me. I reworded it a bit to match the other entries and added some more caveats in https://github.com/electron-userland/electron-packager/commit/8e340292604e7709c7973ea208faaca2d3241198.

PR #414, opened 2016-07-04
Update usage.txt

jaketrent:


malept:
Thanks for the contribution, but the wording of `mas` was deliberate there, as that's the name of one of the possible platform targets. (That's why the other one is called `darwin` instead of `osx`, for example.)

PR #416, opened 2016-07-06
Use fs.rename instead of fs.move when renaming Mac app

joshaber:
Electron apps are big, so any time we can avoid copying the better. Renaming is faster so use it when we safely can.

### DIFF
--- a/mac.js
+++ b/mac.js
@@ -179,14 +179,7 @@ module.exports = {
       operations.push(function (cb) {
         moveHelpers(frameworksPath, opts.name, cb)
       }, function (cb) {
-        var electronAppPath = path.dirname(contentsPath)
-        // If we're not using a temp dir then we know it's safe (and faster!) to
-        // simply rename.
-        if (opts.tmpdir === false) {
-          fs.rename(electronAppPath, finalAppPath, cb)
-        } else {
-          fs.move(electronAppPath, finalAppPath, cb)
-        }
+        fs.rename(path.dirname(contentsPath), finalAppPath, cb)
       })
 
       if ((opts.platform === 'all' || opts.platform === 'mas') && opts['osx-sign'] === undefined) {
--- a/mac.js
+++ b/mac.js
@@ -179,7 +179,14 @@ module.exports = {
       operations.push(function (cb) {
         moveHelpers(frameworksPath, opts.name, cb)
       }, function (cb) {
-        fs.move(path.dirname(contentsPath), finalAppPath, cb)
+        var electronAppPath = path.dirname(contentsPath)
+        // If we're not using a temp dir then we know it's safe (and faster!) to
+        // simply rename.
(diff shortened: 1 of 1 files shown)


malept:
Hmmm. I understand what you're trying to do here, but I don't know whether that's a good assumption to be making. I think I might defer to a developer that's better at the `fs` module than I am.

CC: @jprichardson

malept:
Also, it appears that there is not a test for this change. But I guess hold off on writing that.

develar:
rename works in the electron-builder for a long time, :+1:

malept:
I'm basically worried about using `fs.rename` between different filesystems, volumes, and things like that. There have been several bugs filed in this project similar to that.

develar:
Rename doesn't work across file systems, so, must be used with care. Here checked, that temp die is not used, so, it is ok.

malept:
The temporary directory may not be used, but it doesn't mean that the final `out` param does not point to a different filesystem.

develar:
@malept Sorry, I misunderstood place of this change. It is not common.js, it is mac.js — 

```
const finalAppPath = path.join(tempPath, `${appFilename}.app`)
const contentsPath = path.join(tempPath, 'Electron.app', 'Contents')
```

So, we don't ever need to check is it temp or not. It is always in the same temp fs.

In my fork I use rename in this place in any case — https://github.com/develar/electron-packager/blob/master/mac.js#L164

malept:
Macs can have different filesystems too.

develar:
@malept It is mac.js, not common.js — please see first arg of `path.join` — `tempPath` In this file both vars `finalAppPath` and `contentsPath` are located in the same temp dir. It is not final application out.

malept:
Argh, I need to refactor this file. A 150-line method is a *horrible* idea.

@develar sorry, I'm in the middle of refactoring another part of the code and got confused. You're right.

@joshaber sorry for the confusion, do what @develar suggests and just use `rename` here. No extra test (or `tmpdir` check) necessary.

malept:
Also, either this PR or #415 is going to have a merge conflict, depending on which gets merged first.

joshaber:
Even better!

I'm happy to write a test, but given that there's no externally visible change I'm not sure what we'd test. So long as the existing tests pass!

jprichardson:
As @malept stated:

> I'm basically worried about using fs.rename between different filesystems, volumes, and things like that. There have been several bugs filed in this project similar to that.

is an issue, but one that `fs.move` will handle just fine.

Since @develar stated:

> So, we don't ever need to check is it temp or not. It is always in the same temp fs.

This latest https://github.com/electron-userland/electron-packager/pull/416/commits/9a67489be0c2bcefe3b39bab38ff5bf9e6523827 looks good to me.

malept:
Thanks @joshaber!

joshaber:
Thank you!

PR #417, opened 2016-07-06
Unify asar options

malept:
This is a continuation of #353, but it is rebased against master, merges `asar-options` into `asar`, and adds tests.

It only enhances `asar` and deprecates `asar-unpack` and `asar-unpack-dir`, so only a minor version bump is required.

Closes #353.

## TODO
* [x] Add NEWS entries

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,16 @@
 
 ## Unreleased
 
+### Added
+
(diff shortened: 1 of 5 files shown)


malept:
Assuming the OSX tests ever finish, I'll merge this weekend unless there are any objections.

PR #423, opened 2016-07-11
Add support for Linux ARM

tonychen427:


malept:
Thanks for the contribution, but like #107, this is blocked on https://github.com/electron/electron/issues/2343. Also the tests are probably going to fail.

malept:
Superceded by #474.

PR #438, opened 2016-08-04
add variable to set CFBundleDisplayName

zole40:


malept:
Could you please provide the purpose of this pull request? Also, you're missing documentation.

erkyrath:
Is this addressing https://github.com/electron-userland/electron-packager/issues/387 ?

malept:
If it is, I'd want it to also affect the Windows metadata as well.

zole40:
I wasn't addressing that issue, but this might be a solution for that too - I just needed something similar to version-string.FileDescription for OSX to set the Executable and the Display name independently. Sorry for not adding a description, I'm a bit new to the world of open source

malept:
I'm a little confused, are you suggesting that `version-string.FileDescription` is equivalent to `CFBundleDisplayName` or are you just giving an example?

I would like to have a new parameter that sets CFBundleDisplayName and whatever the equivalent in `version-string` is (I'm not clear on that).

zole40:
Well, my goal was to show the same thing in TaskManager(win) and ActivityMonitor(osx), and on windows the FileDescription is shown while on osx the BundleDisplayName is shown, so in this sense, they are similar. Since I couldn't set the BundleDisplayName, I added a parameter that lets me do that, and posted it in this PR. I hope this answers your question.

malept:
Since there are outstanding technical problems with this pull request, and I don't think this is the correct way to address the problem conceptually, I'm going to close this.

PR #386, opened 2016-06-01
Emit a warning when the wrong variable type is specified for the out parameter

sethlu:
Fixes #385 partially

PR #442, opened 2016-08-06
CLI: ensure --out has either a string or null value

malept:

Yes

**Summarize your changes:**

Closes #386.

This uses a different approach from #386, in that it only handles typecasting for the CLI and not the API. Additionally, it assumes that `--out` without a value is the same as omitting `--out`.

CC: @sethlu

Added NEWS entry

Yes (see `test/cli.js`)

**Did the testsuite pass successfully on your local machine?**

Yes

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


sethlu:
@malept This typecasting method is a neat & clever solution with CLI.

Issue #441, opened 2016-08-05
Fix infer electron test introduced in 7.5.0


malept:
Since #435 was merged, tests for Node 4 in all host platforms intermittently raise an exception at the new infer Electron test.

Examples:

* [Windows](https://ci.appveyor.com/project/electron-userland/electron-packager/build/1.0.181/job/eeydexhmw188r27n#L615)
* [OS X](https://travis-ci.org/electron-userland/electron-packager/jobs/149960137#L275)
* [Linux](https://travis-ci.org/electron-userland/electron-packager/jobs/149908996#L2326)

CC: @zeke

zeke:
ack. investigating.

malept:
I took a couple of minutes to look into it this morning. Here's how I reliably reproduced the problem:

1. Switch to Node 4 environment (I used nodeenv)
2. Take an existing checkout of electron-packager (<= 7.4.0) and run `git pull && npm install`
3. Run `node test/basic.js`, *not* `npm test`

This intermittent error may have actually been an indicator for the problem in #439, but I'll need to see what CI says about that. In the meantime, I committed https://github.com/electron-userland/electron-packager/commit/bcc7cace97e17acbb03cc5d10e16e732532f1687 which at least gives a better exception.

malept:
:disappointed: still happening: https://travis-ci.org/electron-userland/electron-packager/jobs/150300592#L273

I'm dropping it from the 7.5.1 milestone but I'd still like this to get fixed ASAP.

zeke:
Following these steps:

```
n 4.4.7
git checkout tags/v7.4.0
rm -rf node_modules
npm i
node test/basic.js
```

I have run the tests several times and am not able to produce an error.

malept:
If I've learned anything from debugging intermittent test failures in general, it's that running a test in a testsuite vs. a single file has a tendency to be different, if a) your test runner does not isolate your tests properly, or b) you don't write your setup/teardown functions to isolate your tests properly.

My suspicion is that the failure will happen more often when running `npm test` rather than `node test/basic.js`.

malept:
I don't consider this fixed (because it still happens intermittently), so I'm reopening.

malept:
[...]

PR #445, opened 2016-08-08
fix version inference regression

zeke:
Fixes https://github.com/electron-userland/electron-packager/issues/441

This brings `get-package-info` back into the mix, so directories are properly traversed for `package.json` files containing `electron` or `electron-prebuilt`. The work is broken up into two functions now to work around the [`resolve` issue](https://github.com/substack/node-resolve/pull/81) that causes an error to be thrown when a given set of props are not found.

cc @malept @kevinsawicki

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,7 +5,7 @@ const debug = require('debug')('electron-packager')
 const download = require('electron-download')
 const extract = require('extract-zip')
 const fs = require('fs-extra')
-const prop = require('lodash.get')
+const getPackageInfo = require('get-package-info')
 const metadata = require('./package.json')
 const os = require('os')
 const path = require('path')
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -49,28 +50,46 @@ function validateList (list, supported, name) {
...
-  // Infer electron version from package.json
-  var electronVersion = prop(pkg, 'dependencies.electron') || prop(pkg, 'devDependencies.electron')
-  var electronPrebuiltVersion = prop(pkg, 'dependencies.electron-prebuilt') || prop(pkg, 'devDependencies.electron-prebuilt')
-  opts.version = opts.version || electronVersion || electronPrebuiltVersion
-
-  if (!electronVersion && !electronPrebuiltVersion) return cb(null)
+  // Search package.json files to infer name and version from
+  getPackageInfo(props, dir, function (err, result) {
+    if (err) {
+      // `get-package-info` exploded looking for `electron`. Try `electron-prebuilt` instead
```
malept: Wouldn't it also error if both `productName` and `name` were missing?

index.js
```diff
@@ -49,28 +50,46 @@ function validateList (list, supported, name) {
...
-  resolve(packageName, {basedir: dir}, function (err, res, pkg) {
-    if (err) return cb(err)
-    debug('Inferring target Electron version from `' + packageName + '` dependency or devDependency in package.json')
-    opts.version = pkg.version
+function inferNameAndVersionFromInstalled (packageName, opts, result, cb) {
+  if (result.values.productName) {
+    debug('Inferring application name from productName or name in package.json')
+    opts.name = result.values.productName
+  }
+  if (result.values['dependencies.' + packageName]) {
```
malept: Use backticks

index.js
```diff
@@ -49,28 +50,46 @@ function validateList (list, supported, name) {
...
+    resolve(packageName, {
+      basedir: path.dirname(result.source['dependencies.' + packageName].src)
+    }, function (err, res, pkg) {
+      if (err) return cb(err)
+      debug('Inferring target Electron version from ' + packageName + ' dependency or devDependency in package.json')
```
malept: Use backticks (not sure how I missed that last time)

package.json
```diff
@@ -23,6 +23,7 @@
     "electron-osx-sign": "^0.3.0",
     "extract-zip": "^1.0.3",
     "fs-extra": "^0.30.0",
+    "get-package-info": "0.0.3",
     "lodash.get": "^4.4.1",
```
malept: ✂️

index.js
```diff
@@ -5,7 +5,8 @@ const debug = require('debug')('electron-packager')
 const download = require('electron-download')
 const extract = require('extract-zip')
 const fs = require('fs-extra')
-const prop = require('lodash.get')
+const getPackageInfo = require('get-package-info')
+// const prop = require('lodash.get')
```
malept: :scissors:

index.js
rahatarmanahmed: Is `electron` different from `electron-prebuilt`? If they're the same, you can just do `props.push(['dependencies.electron', 'dependencies.electron-prebuilt', 'devDependencies.electron', 'devDependencies.electron-prebuilt'])`, which will resolve to whichever it finds first.

malept: `electron` is preferred over `electron-prebuilt` (the former is the eventual replacement for the latter).

rahatarmanahmed: Then it would be `props.push(['dependencies.electron', 'devDependencies.electron', 'dependencies.electron-prebuilt', 'devDependencies.electron-prebuilt'])`

zeke: Pushing all of them into the same `props` array was what I started with, but then we don't know if the resulting version was derived from `electron` or `electron-prebuilt`. Hence the separation.

zeke:
Now using ES6 string interpolation. I was avoiding it because I thought maybe this project needed to work on older versions of node. But no!

```
  "engines": {
    "node": ">= 4.0"
  }
```

index.js
rahatarmanahmed: Why do we need to know which package it came from?

zeke: Because `resolve` needs a path:

```js
resolve(packageName, {
  basedir: path.dirname(result.source[`dependencies.${packageName}`].src)
}
```

malept:
>  I was avoiding it because I thought maybe this project needed to work on older versions of node. But no!

Yep. And I'm paying for that decision because you can't actually enforce a Node engine version as far as I know. :cry:

index.js
rahatarmanahmed: Oh, d'oh. I forgot about that. In that case it might be better to make a trivial PR to `get-package-info` that can add the property name to `result.source` [here](https://github.com/rahatarmanahmed/get-package-info/blob/master/src/index.js#L36)

rahatarmanahmed: I went ahead and did that myself. Working on a PR for this branch to clean it up accordingly.

zeke:
Thanks for the patch, @rahatarmanahmed. I merged #446. If this passes CI, let's ship it!

malept:
Sounds good. Just needs a NEWS entry and some rebasing.

malept:
There's that error again :disappointed:

zeke:
Added news, squashed. Last CI run passed on all six builds. This is ready. :fingers_crossed:

malept:
Argh https://travis-ci.org/electron-userland/electron-packager/jobs/151351051

zeke:
Rerunning that build. Pretty sure this is just more flakiness.

malept:
Doesn't really fix #441 then, right?

MarshallOfSound:
> Yep. And I'm paying for that decision because you can't actually enforce a Node engine version as far as I know. 

@malept You can be really mean and do something like this in `index.js`

```js
if (parseInt(process.versions.node[0], 10) < 4) {
  throw new BigNastyError('Get a better node version plz')
}
```

malept:
@MarshallOfSound I would, but I think I get a SyntaxError before that can be executed.

zeke:
> Doesn't really fix #441 then, right?

The flakiness on Travis has not been unique to Node 4. I was seeing it on all node versions and platforms. I think this is ready to 🚢

malept:
As far as I can tell, it's always been with that specific test, though.

zeke:
@malept what do you say we ship this since it is definitely an improvement to a known bug in the latest release? We can address the flakiness if it turns out to be an issue.

malept:
Sure. The only thing I object to in this PR is the assertion that it fixes #441.

PR #448, opened 2016-08-11
Add a hook for afterCopy

MarshallOfSound:
* Hook runs once your app dir has been copied to a temp dir
* No hook has no effect (install async callback)
* Generalized the afterExtract test to easily add an afterCopy test

The main use case for this hook is to run `electron-rebuild` or a simply `node-gyp` rebuild script against the **copied** `node_modules` folder.

TLDR: This will make rebuilding `node_modules` for Electron very easy now 👍 (Once I get around to hooking `electron-rebuild` into this)

### DIFF
--- a/common.js
+++ b/common.js
@@ -241,6 +241,14 @@ module.exports = {
         fs.copy(opts.dir, appPath, {filter: userIgnoreFilter(opts), dereference: shouldDeref}, cb)
       },
       function (cb) {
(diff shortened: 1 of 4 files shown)


MarshallOfSound:
>  Coverage increased (+0.04%) to 97.122%

Oooohh yeaaahh.  A whole extra `0.04%` 😆

malept:
This needs a NEWS entry but if I add one, I'll cause a merge conflict with some of the other PRs...so I'll wait.

Issue #325, opened 2016-04-14
Programmatic API options should follow Javascript naming conventions


JonathonAshworth:
In the programmatic API documentation it lists hyphenated properties for the options object such as "app-copyright" and "app-version". These are not valid identifiers and should be camelCased instead, so that string identifiers and array-notation don't have to be used to reference them.

Likewise, the version-string object takes properties that are Capitalized, which is unusual for Javascript objects (though doesn't run into the same problem of having to be defined with a string identifier like the hyphenated ones).

malept:
This is a reasonable request. Something to note for the person who decides to work on this: while camelCased identifiers are idiomatic for JavaScript APIs, they are *not* for the CLI flags. So, in `cli.js` there would need to be either a mapping or a function which consistently converted between the two styles. (The latter of which would obviously need unit tests.)

The other note I would make is that there should be a period of time where both versions of the converted APIs/flags are supported, and the old parameters noted as deprecated in the NEWS file.

malept:
Another thing I am reminded of as I clean up some code is that a lot of the options / sub-options (such as `version-string`) are passthroughs to other modules. So in addition to this issue, I would suggest also filing issues in other repositories, for example, `node-rcedit` (to continue with the example).

malept:
#410 reminded me of this issue. A quick search pointed me at [`camelize`](https://www.npmjs.com/package/camelize).

PR #449, opened 2016-08-11
Infer --app-version from package.json

### LINKED ISSUE #306: `--app-version` should be inferred from package.json when not specified
When `--app-version` is not specified it should just use the `version` field in package.json. It's slightly annoying having to do `--app-version=$npm_package_version` in your npm run script.

> Which version of electron-packager are you using?

6.0.0

zeke:
This infers the `--app-version` from the `package.json` `version` property if it has not been passed in through the command line.

For an app using Electron `1.2.6` with a `package.json` version of `1.1.0`:

Before | After
--- | ---
screen shot 2016-08-11 at 4 08 52 pm | screen shot 2016-08-11 at 4 17 56 pm

cc pairing with @kevinsawicki @jlord 

Fixes #306

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ node_modules
 test/work
 .DS_Store
 .nyc_output
+npm-debug.log
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,7 +4,7 @@
 
 ### Added
 
-* The `package.json` `version` property is the default app version if `--app-version` is unspecified.
+* The `package.json` `version` property is the default app version if `--app-version` is unspecified (#449)
 
 ### Fixed
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Added
+
+* The `package.json` `version` property is the default app version if `--app-version` is unspecified.
+
 ### Fixed
 
 * [CLI] ensure --out has either a string or null value (#442)
--- a/index.js
+++ b/index.js
@@ -51,7 +51,7 @@ function validateList (list, supported, name) {
 function getMetadata (opts, dir, cb) {
   var props = []
   if (!opts.name) props.push(['productName', 'name'])
-  if (!opts['app-version'] || !opts['build-version']) props.push('version')
+  if (!opts['app-version']) props.push('version')
   if (!opts.version) props.push(['dependencies.electron', 'devDependencies.electron'])
 
   // Name and version provided, no need to infer
@@ -80,8 +80,8 @@ function inferNameAndVersionFromInstalled (packageName, opts, result, cb) {
   }
 
   if (result.values.version) {
-    debug('Inferring packageVersion from version in package.json')
-    opts.packageVersion = result.values.version
+    debug('Inferring app-version from version in package.json')
+    opts['app-version'] = result.values.version
   }
 
   if (result.values[`dependencies.${packageName}`]) {
--- a/index.js
+++ b/index.js
@@ -51,8 +51,8 @@ function validateList (list, supported, name) {
 function getMetadata (opts, dir, cb) {
   var props = []
   if (!opts.name) props.push(['productName', 'name'])
-  if (!opts.version) props.push(['dependencies.electron', 'devDependencies.electron'])
   if (!opts['app-version'] || !opts['build-version']) props.push('version')
+  if (!opts.version) props.push(['dependencies.electron', 'devDependencies.electron'])
 
   // Name and version provided, no need to infer
   if (props.length === 0) return cb(null)
--- a/index.js
+++ b/index.js
@@ -48,10 +48,11 @@ function validateList (list, supported, name) {
   return list
 }
 
-function getNameAndVersion (opts, dir, cb) {
+function getMetadata (opts, dir, cb) {
   var props = []
   if (!opts.name) props.push(['productName', 'name'])
   if (!opts.version) props.push(['dependencies.electron', 'devDependencies.electron'])
+  if (!opts['app-version'] || !opts['build-version']) props.push('version')
 
   // Name and version provided, no need to infer
   if (props.length === 0) return cb(null)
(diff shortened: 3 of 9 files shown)


.gitignore
```diff
@@ -6,3 +6,4 @@ node_modules
 test/work
 .DS_Store
 .nyc_output
+basicTest-*/**/*
```
malept: shouldn't that be ignored in `test/work`?

index.js
```diff
@@ -77,6 +78,12 @@ function inferNameAndVersionFromInstalled (packageName, opts, result, cb) {
     debug('Inferring application name from productName or name in package.json')
     opts.name = result.values.productName
   }
+
+  if (result.values.version) {
+    debug('Inferring packageVersion from version in package.json')
+    opts.packageVersion = result.values.version
```
malept: Can we name this `appVersion` in preparation for fixing #325?

kevinsawicki: That would make is a little confusing to me, since that would then have both a dasherized `app-version` and camelcased `appVersion` in the `opts` object.

It would seem better to fix them all as a group via #325

NEWS.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Added
+
+* The `package.json` `version` property is the default app version if `--app-version` is unspecified.
```
malept: Add GitHub PR # so you get credit 😄

malept:
I added the docs-needed tag because I think this needs to be mentioned in at least `docs/api.md`. Not sure if it needs to be mentioned in the readme.

jlord:
@malept Oh yeah, of course! Our oversight. I'll add that 👍

malept:
The consistent CI failures on OSX remind me why I didn't implement this feature myself - I couldn't figure out a way to get Wine to cooperate on Travis CI's OSX workers, so I just disabled those tests. Might have to do the same thing here.

test/ci/before_install.sh
```diff
@@ -24,5 +24,6 @@ case "$TRAVIS_OS_NAME" in
       -in codesign.csr -out codesign.cer
     openssl pkcs12 -export -in codesign.cer -inkey codesign.key -out codesign.p12 -password pass:12345
     security import codesign.p12 -k ~/Library/Keychains/login.keychain -P 12345 -T /usr/bin/codesign
+    brew install wine
```
malept: FYI, the last time I tried this I had to do some weird things so that the OSX CI builds didn't take forever to run and used a relatively recent version of Wine: https://github.com/electron-userland/electron-packager/blob/v7.1.0/test/ci/before_install.sh#L6-L11

zeke: Why did you end up taking the brew installation out?

malept: Tests involving rcedit kept timing out.

malept:
I think I'm going to drop this for the 7.6.0 release and save it for a future non-patch release (which will cause a merge conflict in NEWS.md).

test/ci/before_install.sh
```diff
@@ -24,5 +24,6 @@ case "$TRAVIS_OS_NAME" in
       -in codesign.csr -out codesign.cer
     openssl pkcs12 -export -in codesign.cer -inkey codesign.key -out codesign.p12 -password pass:12345
     security import codesign.p12 -k ~/Library/Keychains/login.keychain -P 12345 -T /usr/bin/codesign
+    npm install wine-darwin@1.9.16
```
malept: NPM can install *everything*, I guess...

test/util.js
```diff
@@ -75,7 +75,14 @@ exports.setup = function setup () {
     fs.mkdirp(WORK_CWD, function (err) {
       if (err) t.end(err)
       process.chdir(WORK_CWD)
-      t.end()
+
+      if (process.platform === 'darwin' && process.env.TRAVIS) {
+        fs.remove(path.join(process.env.HOME, '.wine'), function () {
```
malept: This seems like it would make everything slower.

kevinsawicki: Yeah, this was just an attempt to see what could be causing it to fail on CI, will revert after investigating more.

kevinsawicki:
Got specs passing on Travis CI using `wine`, https://travis-ci.org/electron-userland/electron-packager/builds/153046665

malept:
Nice! The question becomes, is it consistent, or will tests time out sometimes? The inconsistent timeouts were why I had to stop installing wine on Travis OSX in the first place.

kevinsawicki:
> The question becomes, is it consistent, or will tests time out sometimes?

Yeah, since this PR causes way more tests to call `rcedit` via `wine` and the build passed on all 3 macs in the matrix configuration, I'm optimistic that things will be consistent given the new way wine is being installed and configured.

I haven't seen any failures on https://travis-ci.org/kevinsawicki/wine-darwin once the `WINEDLLOVERRIDES` env var was set which is also set now in this PR.

malept:
Not that I want to drag this PR out any longer, but should we have a test for the case where no `app-version` is specified *and* (for some reason) there is no `version` in any of the `package.json` files found?

malept:
I asked:

> [...] should we have a test [...]

We talked about it this morning - I believe where we ended up is that it would be nice but it would take a lot of effort to handle this case, because you'd have to either:

* Copy a fixture to a temporary directory (so that the electron-packager `package.json` doesn't get detected by `get-package-info`) - and even then, there could be inconsistent failures on dev machines where someone puts a `package.json` in their root temporary directory
* Refactor that code to stop at the first `package.json` - I'm generally in favor of this method, but the concern was that it would break existing code (even though that wasn't an explicit feature)

@kevinsawicki please correct me if I'm misremembering.

My inclination is for the second option to happen, but not in this PR. It would also obviously be a breaking change that requires a major version bump.

zeke:
🎉

malept:
Feel free to release 7.7.0 (or I can do it sometime after this afternoon). I need to add a couple of steps to the release docs for after running `git push`:

* Create a new GitHub release from the pushed tag with the contents of `NEWS` for that version
* Close the milestone associated with the version if one is open

PR #452, opened 2016-08-12
Handle errors setting up tests

kevinsawicki:
Previously, if `npm install` on the fixtures failed or if Electron download failed, the tests would not fail.

This pull request adds error handling when either of those fails.

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -38,7 +38,12 @@ function npmInstallforFixtures () {
 
 let setupFuncs = preDownloadElectron().concat(npmInstallforFixtures())
 
-series(setupFuncs, () => {
+series(setupFuncs, (error) => {
+  if (error) {
+    console.error(error.stack || error)
+    return process.exit(1)
+  }
+
   console.log('Running tests...')
   require('./basic')
   require('./asar')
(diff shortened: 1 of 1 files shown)


malept:
That is a bit odd conceptually, but if it works, it works.

kevinsawicki:
> That is a bit odd conceptually, but if it works, it works.

Yeah, agreed, it seemed to kind of mesh with the setup/teardown as tap tests used elsewhere.

It could just handle the errors explicitly and do a manual `process.exit` when downloading Electron or npm installing fails if you think that is a better approach.

malept:
> It could just handle the errors explicitly and do a manual `process.exit` when downloading Electron or npm installing fails if you think that is a better approach.

I like that better.

kevinsawicki:
> I like that better.

Updated 👍

malept:
Hmm, this seems bad: https://travis-ci.org/electron-userland/electron-packager/jobs/151845468#L2066-L2083

I wonder if it has something to do with #441.

zeke:
We put `electron-prebuilt` in `dependencies` and `devDependencies` in that fixture to ensure that `electron` will always take precedence, regardless of how `electron-prebuilt` is installed. But if that is causing issues we can't get around, we should probably just use `electron-prebuilt` once in the `package.json` file.

malept:
I think that will be fixed if we run `npm install --no-bin-links`.

kevinsawicki:
@malept Looks like the latest round of builds all pass, 🚢 ❔

PR #459, opened 2016-08-18
Disallow setting osx-sign.binaries

### LINKED ISSUE #285: Pass binaries option to electron-osx-sign

malept:

Yes

**Summarize your changes:**

Removes the suboption `osx-sign.binaries` if it is set and print a warning to that effect. See discussion in #285 and #286 for context.

CC: @sethlu 

Closes #285.

Not necessary, the docs specify which suboptions are supported.

Yes

Yes

### DIFF
--- a/test/mac.js
+++ b/test/mac.js
@@ -572,6 +572,13 @@ module.exports = function (baseOpts) {
     t.end()
   })
 
+  test('osx-sign argument test: binaries not set', (t) => {
+    let args = {binaries: ['binary1', 'binary2']}
+    let signOpts = mac.createSignOpts(args, 'darwin', 'out')
+    t.same(signOpts, {app: 'out', platform: 'darwin'})
+    t.end()
+  })
+
   util.setup()
   test('codesign test', function (t) {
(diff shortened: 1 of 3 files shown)


test/mac.js
```diff
@@ -545,6 +545,14 @@ module.exports = function (baseOpts) {
     t.end()
   })
 
+  test('osx-sign argument test: binaries not set', (t) => {
+    let args = {binaries: ['binary1', 'binary2']}
+    let signOpts = mac.createSignOpts(args, 'darwin', 'out')
+    console.log(signOpts)
```
kevinsawicki: :fire:

malept: :facepalm:

PR #464, opened 2016-08-22
Add default values to arch and platform

### LINKED ISSUE #36: `--arch` should default to host arch
Good defaults again. Less config.

malept:

Yes

**Summarize your changes:**

The defaults are the host's arch and platform values.

Fixes #36.

Updated NEWS, <ins>readme</ins>, API, & CLI docs. <del>I probably need to update the readme as well (have not done that yet).</del>

Changed the defaults test to test the new platform/arch defaults.

Yes

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,11 @@
 
 * `win32metadata` option (#331, #463)
 
+### Changed
+
+* Default the `platform` option to the host platform (#464)
+* Default the `arch` option to the host arch (#36, #464)
+
 ### Fixed
 
 * Allow scoped package names as Electron app names - invalid characters are replaced with
--- a/cli.js
+++ b/cli.js
@@ -18,7 +18,7 @@ var usage = fs.readFileSync(path.join(__dirname, 'usage.txt')).toString()
 
 var args = common.parseCLIArgs(process.argv.slice(2))
 
-if (!args.dir || (!args.all && (!args.platform || !args.arch))) {
+if (!args.dir) {
   // temporary fix for https://github.com/nodejs/node/issues/6456
   if (process.stderr._handle && process.stderr._handle.setBlocking) {
     process.stderr._handle.setBlocking(true)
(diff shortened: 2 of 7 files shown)


malept:
@kevinsawicki @jlord @zeke any comments/concerns you have on either the concept or the changes would be much appreciated. This one is likely going to be a fairly controversial change.

kevinsawicki:
>  any comments/concerns you have on either the concept or the changes would be much appreciated.

My instinct is to keep 32-bit as the default arch instead of the host arch (which is presumably 64-bit since developers commonly use 64-bit machines). When unspecified, I'm assuming the person wants a package that will work on the most machines possible, and 32-bit will work more places than 64-bit.

For platform though, the host platform sounds like a great default that people will expect 👍

malept:
My  rationale for making both defaults the same as the host is that it is easier for the developer to test. (I should have put that in the PR summary.)

I believe that Windows 32-bit executables will run on 64-bit machines (maybe? I could be wrong) but in order for Linux 32-bit executables to run, the whole 32-bit glibc/etc stack also needs to be installed (e.g., `dpkg --add-architecture i386`), unless for some reason Electron is statically linked to musl or similar :smile:

kevinsawicki:
> I believe that Windows 32-bit executables will run on 64-bit machines 

Yup, that is my understanding as well.

> but in order for Linux 32-bit executables to run, the whole 32-bit glibc/etc stack also needs to be installed (e.g., dpkg --add-architecture i386)

Oh, interesting, didn't realize this.

Since it seems like the better default on Linux is the host arch, and their is no 32-bit on macOS, the host arch is the better default.

One question, if I am on 32-bit Linux and Windows, and I request to build for macOS with no arch specified, will it still build a 64-bit version of mac since their is no 32-bit, or would it error about 32-bit mac being unavailable?

malept:
> One question, if I am on 32-bit Linux and Windows, and I request to build for macOS with no arch specified, will it still build a 64-bit version of mac since their is no 32-bit, or would it error about 32-bit mac being unavailable?

It will currently error. It would also error if #106 is fixed and you only specify `--arch=armv7l` on OSX/Windows.

zeke:
Nice! Are you planning to land this in 8.0.0?

malept:
Yup!

PR #472, opened 2016-08-26
Default the prune option to true

### LINKED ISSUE #235: Not currently possible to ignore devDependencies with npm3

malept:

Yes

**Summarize your changes:**

In both the CLI and API, `prune` defaults to `true` (the CLI is just more explicit about it in the code).

@zeke I know you assigned yourself to #35 but I had some time the other day to work on this.

Addresses #35. (Obviously, not the `asar` part.)
Fixes #235.

The CLI docs now mentions `--no-prune` instead of `--prune`. Also updated NEWS, the readme, and the API docs.

Updated the defaults test, added a test for prune=false.

Yes
## TODO

* [x] truthy or undefined *([ref](#discussion_r76352707))*
* [x] `--no-prune` should go after `ignore` now to keep things alphabetical *([ref](#discussion_r76538010))*

### DIFF
--- a/common.js
+++ b/common.js
@@ -195,7 +195,7 @@ module.exports = {
     // * Creates temporary directory
     // * Copies template into temporary directory
     // * Copies user's app into temporary directory
-    // * Prunes non-production node_modules (if opts.prune is not false)
+    // * Prunes non-production node_modules (if opts.prune is either truthy or undefined)
     // * Creates an asar (if opts.asar is set)
 
     var tempPath
--- a/common.js
+++ b/common.js
@@ -24,7 +24,8 @@ function parseCLIArgs (argv) {
     ],
     default: {
       'deref-symlinks': true,
-      'download.strictSSL': true
+      'download.strictSSL': true,
+      prune: true
     },
     string: [
       'out'
@@ -194,7 +195,7 @@ module.exports = {
     // * Creates temporary directory
     // * Copies template into temporary directory
     // * Copies user's app into temporary directory
-    // * Prunes non-production node_modules (if opts.prune is set)
+    // * Prunes non-production node_modules (if opts.prune is not false)
     // * Creates an asar (if opts.asar is set)
 
     var tempPath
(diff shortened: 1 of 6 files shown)


common.js
```diff
@@ -194,7 +195,7 @@ module.exports = {
     // * Creates temporary directory
     // * Copies template into temporary directory
     // * Copies user's app into temporary directory
-    // * Prunes non-production node_modules (if opts.prune is set)
+    // * Prunes non-production node_modules (if opts.prune is not false)
```
zeke: truthy?

malept: Pedantically speaking, either truthy or `undefined`.

usage.txt
```diff
@@ -48,7 +48,7 @@ overwrite          if output directory for a platform already exists, replaces i
                    skipping it
 platform           all, or one or more of: darwin, linux, mas, win32 (comma-delimited if multiple).
                    Defaults to the host platform
-prune              runs `npm prune --production` on the app
+no-prune           do not run `npm prune --production` on the app
```
kevinsawicki: Do you think this should this go after `ignore` now to keep things alphabetical?

malept: Yep!

zeke:
How come CI is not running on this PR?

malept:
It did in the previous commit (note the checkmark), I skipped it in the latest one because it was just a doc change. That's some interesting UX implications in the GitHub PR UI :smiley:

zeke:
Cool. Is this ready to ship, then?

malept:
Yup!

PR #479, opened 2016-09-01
Replace newlines with spaces

StefMa:
The issue template contains  some unnecessary new lines.
I've replaced it with spaces 🎉


malept:
Thanks for the pull request, but the newlines are intentional. In this repository, I try to make Markdown lines wrap at around 100 characters.

Issue #461, opened 2016-08-20
Simplify package.json inference


zeke:
Following up on https://github.com/electron-userland/electron-packager/pull/449#issuecomment-240865736

Currently there's a bit of code in `index.js` that uses `get-package-info` to look for package.json files from which to infer default values for electron version, product name, etc. This technique of recursing up directories can produce unpredictable results depending on the directory structure of a given user.

As @malept said:

> Not that I want to drag this PR out any longer, but should we have a test for the case where no app-version is specified and (for some reason) there is no version in any of the package.json files found?

Looking for the *first* parent package.json file instead of *all* parent package.json files will enable us to reliably test that scenario. Writing up this issue because I came across a battle-tested package that does what we want: https://www.npmjs.com/package/find-root

malept:
One of the things we should consider when we redo the `getMetadata` method (and related) is to move it to `common.js`, so it can be unit tested without needing to run the whole `packager` workflow.

rahatarmanahmed:
Have there been unpredictable results? AFAIK the only issue I've heard of is the unhelpful error that is returned in that case where the version can't be found in any package.json files. That is something that can be addressed in https://github.com/rahatarmanahmed/get-package-info/issues/21

malept:
@rahatarmanahmed there have been odd results on Windows, see #497.

rahatarmanahmed:
@malept I've just added debug messages so we can use  `DEBUG=get-package-info` for more info on what files are being checked. It's a patch release so should be included in new installs.

I'm working on knocking this issue out by using `read-pkg-up` package instead of `resolve`.

rahatarmanahmed:
Ok, published `1.0.0` of `get-package-info`, which now returns a better error that lets us know when props can't be found.

From the README:

> If all the properties cannot be found in parent package.json files, then getPackageInfo() will reject it's promise (or callback with err argument) with an Error. err.missingProps will have an array of the properties that it could not find, and err.result will contain all the props that were found.
>
> If any other error occurs(like I/O or runtime errors), getPackageInfo() will reject with that error itself.

I'll work on a PR using this new info later tonight.

rahatarmanahmed:
@zeke I'm noticing that my changes back from https://github.com/electron-userland/electron-packager/pull/446/files removing the 2nd call to `getPackageInfo` didn't end up in master. Was it lost in the rebase or was there a reason to keep the 2nd call?

zeke:
[...]

Issue #154, opened 2015-10-12
Rename --version to --electron-version


malept:
There seem to be a disproportionate number of issues created over the confusion around what `--version` actually means. (See: #137, #140, #153.) I propose that the parameter be renamed so that it's obvious as to what the flag's value refers to: the version of Electron, not the version of the application.

This would obviously be a breaking change, requiring a major version bump.

maxogden:
I'd rather try and fix this with docs first. The readme formatting could use some love, e.g. the usage section is currently a big wall of unicode text

malept:
I figured as much. I just wanted to get this idea out there and documented.

I'll take a crack at making the README friendlier tonight, if I have some time.

Closing for now.

malept:
I'm reopening this. No amount of docs has fixed this so far. In my opinion, this option violates [the principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

PR #480, opened 2016-09-02
Add check for additional properties for better error reporting

kokarn:
**Summarize your changes:**
Minor logging update to correctly display what properties we look for when building.

Probably related to https://github.com/electron-userland/electron-packager/issues/476

EDIT: Sorry about the spam :/

### DIFF
--- a/index.js
+++ b/index.js
@@ -63,7 +63,7 @@ function getMetadata (opts, dir, cb) {
     if (err) {
       // `get-package-info` exploded looking for `electron`. Try `electron-prebuilt` instead
       props.pop()
-      props.push(['dependencies.electron-prebuilt', 'devDependencies.electron-prebuilt'])
+      props.push(['dependencies.electron', 'devDependencies.electron', 'dependencies.electron-prebuilt', 'devDependencies.electron-prebuilt'])
       getPackageInfo(props, dir, function (err, result) {
         if (err) return cb(err)
         return inferNameAndVersionFromInstalled('electron-prebuilt', opts, result, cb)


malept:
Thank you for the pull request, but could you be more specific about the impact of the change? How does this make for better error reporting? Please give an example of how the logging changes.

malept:
_Additionally, in the future, making sure that all of the questions in the pull request template are answered would cut down on the amount of notification spam generated by your pull request._

kokarn:
So sorry about the questions! Was thinking because they didn't really add any value it was safe to remove.

Currently when packaging from `C:\repos\my-app` with an invalid package.json missing electron & electron-prebuild as (dev)dependency this is the output
```sh
> electron-packager . --platform=win32 --arch=x64

Unable to determine application name or Electron version. Please specify an application name and Electron version.

For more infomation, please see
https://github.com/electron-userland/electron-packager/blob/master/docs/api.md#name or
https://github.com/electron-userland/electron-packager/blob/master/docs/api.md#version

Cannot find module 'package.json' from 'C:\repos'(Properties not found yet: dependencies.electron-prebuilt,devDependencies.electron-prebuilt)
```

With the changes this is the output
```
> electron-packager . --platform=win32 --arch=x64

Unable to determine application name or Electron version. Please specify an application name and Electron version.

For more infomation, please see
https://github.com/electron-userland/electron-packager/blob/master/docs/api.md#name or
https://github.com/electron-userland/electron-packager/blob/master/docs/api.md#version

Cannot find module 'package.json' from 'C:\repos'(Properties not found yet: dependencies.electron,devDependencies.electron,dependencies.electron-prebuilt,devDependencies.electron-prebuilt)
```

I'm aware it's a very minimal change but I believe it would help to remove some confusion for first time adopters as it did for me.

PS: I've filed a separate bug report for the incorrect path and ugliness of the output message

malept:
Right now I don't think this is the right solution to the problem. Adding those entries causes unnecessary checking via the `get-package-info` module. I believe that the correct fix is the one described in https://github.com/electron-userland/electron-packager/issues/461.

Does the documentation in https://github.com/electron-userland/electron-packager/blob/master/docs/api.md#version need to be updated to be more clear?

kokarn:
Yeah you are probably correct in that.

Hmm, probably not. 

For a little while I mistook the version in package.json as the "version" it was referring to but that's probably just me being a bit hasty.

Let's hope for #461 instead.

malept:
> For a little while I mistook the version in package.json as the "version" it was referring to

This kind of reaction is why I never liked the electron version option named "version". I think it should be named `electronVersion` (or `--electron-version` on the command line).

kokarn:
It's kinda of weird, yeah. 

First impressions is a bit all over the place. 

The functionality is amazing but the mix & match between using custom props (productName), props with the same name (version) and default props (dependencies/devDependencies) makes it a bit confusing at first.

I think id actually prefer a custom dotfile just for electron-packager or something like that just to make it 100% clear.

malept:
FWIW I'd like to move `productName` to a more standard place like https://docs.npmjs.com/files/package.json#config. I'd also like the ability to configure all (most?) of Electron Packager via the `config` section of `package.json`, but that would require more thought.

malept:
> This kind of reaction is why I never liked the electron version option named "version". I think it should be named `electronVersion` (or `--electron-version` on the command line).

I've reopened #154 as a result.

kokarn:
Great! Never heard of "Principle of least astonishment" but i totally agree with that sentiment.

PR #485, opened 2016-09-05
Updated inferNameAndVersionFromInstalled

Shinrai:
Now properly resolves devDependencies as documented in the debug code.
Previously it would not look at devDependencies.

### DIFF
--- a/index.js
+++ b/index.js
@@ -54,10 +54,15 @@ function inferNameAndVersionFromInstalled (packageName, opts, result, cb) {
     debug('Inferring app-version from version in package.json')
     opts['app-version'] = result.values.version
   }
-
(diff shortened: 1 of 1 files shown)


malept:
Thanks for filing a pull request. Per our [CONTRIBUTING instructions](https://github.com/electron-userland/electron-packager/blob/master/CONTRIBUTING.md#filing-pull-requests):

> Unless it's impractical, please write tests for your changes. This will help us so that we can spot regressions much easier.

malept:
Also, it appears that the changes don't conform to our coding standard (also documented in the CONTRIBUTING instructions). In the future, this can be avoided by running the tests locally before submitting a pull request.

PR #501, opened 2016-09-29
Default build version to app version  on Windows

### LINKED ISSUE #498: Build version should default to app version on Windows
**Please describe your issue:**

On Mac, both plist version keys are set to the `app-version` option when present, https://github.com/electron-userland/electron-packager/blob/9134f7be4678fdf0144bbb6b00f2da4756566415/mac.js#L109-L111 so you can just set the `app-version` option when you want the app version and build version to have the same value, but you can still override `build-version` when needed.

This is not happening on Windows, https://github.com/electron-userland/electron-packager/blob/9134f7be4678fdf0144bbb6b00f2da4756566415/win32.js#L36-L42

Perhaps on Windows, the `file-version` and `product-version` should both default to `app-version` initially but `file-version` should still be overrideable by `build-version` when it is present.

This would make the `app-version` defaulting to the `package.json` `version` value introduced in #449 be more consistent on Windows and Mac.

kevinsawicki:

Totes

**Summarize your changes:**

Default the `build-version` option to the `app-version` when it is unspecified. This makes the Windows behavior consistent with the macOS defaulting behavior introduced in  #449.

Closes #498 

Yup

Yes

Yes

### DIFF
--- a/test/win32.js
+++ b/test/win32.js
@@ -56,11 +56,11 @@ function setFileVersionTest (buildVersion) {
     'build-version': buildVersion
   }
 
-  return generateVersionStringTest(['FileVersion', 'ProductVersion'],
+  return generateVersionStringTest(['ProductVersion', 'FileVersion'],
                                    opts,
-                                   [buildVersion, '4.99.101.0'],
-                                   ['File version should match build version',
-                                    'Product version should match package.json version'])
+                                   ['4.99.101.0', buildVersion],
+                                   ['Product version should match package.json version',
+                                    'File version should match build version'])
 }
 
 function setProductVersionTest (appVersion) {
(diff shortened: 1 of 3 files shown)


test/win32.js
```diff
@@ -155,6 +171,7 @@ util.packagerTest('win32 executable name is based on sanitized app name', (t) =>
 
 util.packagerTest('win32 build version sets FileVersion test', setFileVersionTest('2.3.4.5'))
 util.packagerTest('win32 app version sets ProductVersion test', setProductVersionTest('5.4.3.2'))
+util.packagerTest('win32 app/build version default to package.json test', defaultProductAndFileVersionToPackageVersionTest())
```
malept: app/build versions

PR #505, opened 2016-10-13
Update get-package-info

rahatarmanahmed:
**Have you read the [section in CONTRIBUTING.md about pull requests](https://github.com/electron-userland/electron-packager/blob/master/CONTRIBUTING.md#filing-pull-requests)?** Yes

**Summarize your changes:**

As discussed in #461, this updates `get-package-info` to v1.0.0, which includes more helpful errors when failing to find all info requested. This lets us give better error messages, as well as ignore missing props for optional options.

**Are your changes appropriately documented?** N/A

**Do your changes have sufficient test coverage?** Not sure yet.

**Does the testsuite pass successfully on your local machine?** Yes

### DIFF
--- a/index.js
+++ b/index.js
@@ -83,19 +83,23 @@ function getMetadata (opts, dir, cb) {
       opts['app-version'] = result.values.version
     }
 
-    if (result.values[`dependencies.electron`]) {
-      var packageName = result.source[`dependencies.electron`].prop.split('.')[1]
-      resolve(packageName, {
-        basedir: path.dirname(result.source[`dependencies.electron`].src)
-      }, function (err, res, pkg) {
-        if (err) return cb(err)
-        debug(`Inferring target Electron version from ${result.source['dependencies.electron'].prop} in ${result.source['dependencies.electron'].src}`)
-        opts.version = pkg.version
-        return cb(null)
-      })
+    if (result.values['dependencies.electron']) {
+      let {prop, src} = result.source['dependencies.electron']
+      return getVersion(opts, prop.split('.')[1], src, cb)
     } else {
       return cb(null)
     }
+  })
+}
+
+function getVersion (opts, packageName, src, cb) {
+  resolve(packageName, {
+    basedir: path.dirname(src)
+  }, (err, res, pkg) => {
+    if (err) return cb(err)
+    debug(`Inferring target Electron version from ${packageName} in ${src}`)
+    opts.version = pkg.version
+    return cb(null)
   })
 }
 
--- a/index.js
+++ b/index.js
@@ -67,8 +67,9 @@ function getMetadata (opts, dir, cb) {
     }
 
     if (result.values['dependencies.electron']) {
-      let {prop, src} = result.source['dependencies.electron']
-      return getVersion(opts, prop.split('.')[1], src, cb)
+      let prop = result.source['dependencies.electron'].prop.split('.')[1]
+      let src = result.source['dependencies.electron'].src
+      return getVersion(opts, prop, src, cb)
     } else {
       return cb(null)
     }
--- a/index.js
+++ b/index.js
@@ -23,50 +23,77 @@ function getMetadata (opts, dir, cb) {
   var props = []
   if (!opts.name) props.push(['productName', 'name'])
   if (!opts['app-version']) props.push('version')
-  if (!opts.version) props.push(['dependencies.electron', 'devDependencies.electron'])
+  if (!opts.version) {
+    props.push([
+      'dependencies.electron',
+      'devDependencies.electron',
+      'dependencies.electron-prebuilt',
+      'devDependencies.electron-prebuilt'
+    ])
+  }
 
   // Name and version provided, no need to infer
   if (props.length === 0) return cb(null)
 
   // Search package.json files to infer name and version from
-  getPackageInfo(props, dir, function (err, result) {
-    if (err) {
-      // `get-package-info` exploded looking for `electron`. Try `electron-prebuilt` instead
-      props.pop()
-      props.push(['dependencies.electron-prebuilt', 'devDependencies.electron-prebuilt'])
-      getPackageInfo(props, dir, function (err, result) {
+  getPackageInfo(props, dir, (err, result) => {
+    if (err && err.missingProps) {
+      var requiredProps = ['productName', 'dependencies.electron']
+      var missingProps = err.missingProps.map(prop => {
+        return Array.isArray(prop) ? prop[0] : prop
+      })
+
+      // Callback w/ error if there are required props that are missing
+      if (missingProps.filter(prop => requiredProps.find(reqProp => prop === reqProp)).length !== 0) {
+        var messages = missingProps.map(function (missingProp) {
+          missingProp = Array.isArray(missingProp) ? missingProp[0] : missingProp
+          if (missingProp === 'productName') {
+            return 'Unable to determine application name. Please specify an application name\n\n' +
+              'For more information, please see\n' +
+              'https://github.com/electron-userland/electron-packager/blob/master/docs/api.md#name\n'
+          }
+
+          if (missingProp === 'dependencies.electron') {
+            return 'Unable to determine Electron version. Please specify an Electron version\n\n' +
+              'For more information, please see\n' +
+              'https://github.com/electron-userland/electron-packager/blob/master/docs/api.md#version\n'
+          }
+        })
+
+        err.message = messages.join('\n') + '\n' + err.message
+        return cb(err)
+      } else {
+        // Missing props not required, can continue w/ partial result
+        result = err.result
+      }
+    } else if (err) {
+      return cb(err)
+    }
+
+    if (result.values.productName) {
+      debug('Inferring application name from productName or name in package.json')
+      opts.name = result.values.productName
+    }
+
+    if (result.values.version) {
+      debug('Inferring app-version from version in package.json')
+      opts['app-version'] = result.values.version
+    }
+
+    if (result.values[`dependencies.electron`]) {
+      var packageName = result.source[`dependencies.electron`].prop.split('.')[1]
+      resolve(packageName, {
+        basedir: path.dirname(result.source[`dependencies.electron`].src)
+      }, function (err, res, pkg) {
         if (err) return cb(err)
-        return inferNameAndVersionFromInstalled('electron-prebuilt', opts, result, cb)
+        debug(`Inferring target Electron version from ${packageName} dependency or devDependency in package.json`)
+        opts.version = pkg.version
+        return cb(null)
       })
     } else {
-      return inferNameAndVersionFromInstalled('electron', opts, result, cb)
+      return cb(null)
     }
   })
-}
-
-function inferNameAndVersionFromInstalled (packageName, opts, result, cb) {
-  if (result.values.productName) {
-    debug('Inferring application name from productName or name in package.json')
-    opts.name = result.values.productName
-  }
-
-  if (result.values.version) {
-    debug('Inferring app-version from version in package.json')
-    opts['app-version'] = result.values.version
-  }
-
-  if (result.values[`dependencies.${packageName}`]) {
-    resolve(packageName, {
-      basedir: path.dirname(result.source[`dependencies.${packageName}`].src)
-    }, function (err, res, pkg) {
-      if (err) return cb(err)
-      debug(`Inferring target Electron version from ${packageName} dependency or devDependency in package.json`)
-      opts.version = pkg.version
-      return cb(null)
-    })
-  } else {
-    return cb(null)
-  }
 }
 
 function createSeries (opts, archs, platforms) {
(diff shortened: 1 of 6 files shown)


index.js
```diff
@@ -23,50 +23,77 @@ function getMetadata (opts, dir, cb) {
...
+          }
+
+          if (missingProp === 'dependencies.electron') {
+            return 'Unable to determine Electron version. Please specify an Electron version\n\n' +
+              'For more information, please see\n' +
+              'https://github.com/electron-userland/electron-packager/blob/master/docs/api.md#version\n'
+          }
+        })
+
+        err.message = messages.join('\n') + '\n' + err.message
```
rahatarmanahmed: May want to hide original error message, since the message we add should be descriptive enough without it?

index.js
```diff
@@ -23,50 +23,77 @@ function getMetadata (opts, dir, cb) {
...
+        return cb(err)
+      } else {
+        // Missing props not required, can continue w/ partial result
```
rahatarmanahmed: Assuming `app-version` is not a required option.

index.js
```diff
@@ -23,50 +23,77 @@ function getMetadata (opts, dir, cb) {
...
+        return Array.isArray(prop) ? prop[0] : prop
       })
-    } else {
-      return inferNameAndVersionFromInstalled('electron', opts, result, cb)
+
+      // Callback w/ error if there are required props that are missing
+      if (missingProps.filter(prop => requiredProps.find(reqProp => prop === reqProp)).length !== 0) {
+        var messages = missingProps.map(function (missingProp) {
+          missingProp = Array.isArray(missingProp) ? missingProp[0] : missingProp
+          if (missingProp === 'productName') {
```
malept: Looks like you need to add a test for this conditional.

index.js
```diff
@@ -23,50 +23,77 @@ function getMetadata (opts, dir, cb) {
...
+            return 'Unable to determine application name. Please specify an application name\n\n' +
+              'For more information, please see\n' +
+              'https://github.com/electron-userland/electron-packager/blob/master/docs/api.md#name\n'
+          }
+
+          if (missingProp === 'dependencies.electron') {
+            return 'Unable to determine Electron version. Please specify an Electron version\n\n' +
+              'For more information, please see\n' +
+              'https://github.com/electron-userland/electron-packager/blob/master/docs/api.md#version\n'
```
malept: It would be nice if these error messages could be DRY'd up somehow.

index.js
malept: `app-version` isn't required. From `docs/api.md`:

> If neither [`app-version` option nor `version` from `package.json`] are provided, the version of Electron will be used.

index.js
malept: > original error message

The one that's sent by `get-package-info`? If so, perhaps it should be printed in a `debug()` message.

index.js
```diff
@@ -23,50 +23,77 @@ function getMetadata (opts, dir, cb) {
...
-      if (err) return cb(err)
-      debug(`Inferring target Electron version from ${packageName} dependency or devDependency in package.json`)
-      opts.version = pkg.version
+    if (result.values[`dependencies.electron`]) {
+      var packageName = result.source[`dependencies.electron`].prop.split('.')[1]
+      resolve(packageName, {
+        basedir: path.dirname(result.source[`dependencies.electron`].src)
+      }, function (err, res, pkg) {
+        if (err) return cb(err)
+        debug(`Inferring target Electron version from ${packageName} dependency or devDependency in package.json`)
```
malept: Can we make this debug statement more specific now with `result.source`?

index.js
```diff
@@ -23,50 +23,77 @@ function getMetadata (opts, dir, cb) {
...
-  })
-}
 
-function inferNameAndVersionFromInstalled (packageName, opts, result, cb) {
-  if (result.values.productName) {
-    debug('Inferring application name from productName or name in package.json')
-    opts.name = result.values.productName
-  }
+    if (result.values.productName) {
+      debug('Inferring application name from productName or name in package.json')
```
malept: Can we make this debug statement more specific now with `result.source`?

index.js
```diff
@@ -23,50 +23,77 @@ function getMetadata (opts, dir, cb) {
...
-      // `get-package-info` exploded looking for `electron`. Try `electron-prebuilt` instead
-      props.pop()
-      props.push(['dependencies.electron-prebuilt', 'devDependencies.electron-prebuilt'])
-      getPackageInfo(props, dir, function (err, result) {
-        if (err) return cb(err)
-        return inferNameAndVersionFromInstalled('electron-prebuilt', opts, result, cb)
+  getPackageInfo(props, dir, (err, result) => {
+    if (err && err.missingProps) {
+      var requiredProps = ['productName', 'dependencies.electron']
+      var missingProps = err.missingProps.map(prop => {
```
malept: `let`

index.js
```diff
@@ -23,50 +23,77 @@ function getMetadata (opts, dir, cb) {
...
+        return Array.isArray(prop) ? prop[0] : prop
       })
-    } else {
-      return inferNameAndVersionFromInstalled('electron', opts, result, cb)
+
+      // Callback w/ error if there are required props that are missing
+      if (missingProps.filter(prop => requiredProps.find(reqProp => prop === reqProp)).length !== 0) {
+        var messages = missingProps.map(function (missingProp) {
```
malept: Use `let` + fat arrow syntax here?

index.js
```diff
@@ -23,50 +23,77 @@ function getMetadata (opts, dir, cb) {
...
+          missingProp = Array.isArray(missingProp) ? missingProp[0] : missingProp
```
malept: Didn't you already do this earlier when declaring `missingProps`?

index.js
```diff
@@ -23,50 +23,77 @@ function getMetadata (opts, dir, cb) {
...
+      debug('Inferring application name from productName or name in package.json')
+      opts.name = result.values.productName
+    }
 
-  if (result.values.version) {
-    debug('Inferring app-version from version in package.json')
-    opts['app-version'] = result.values.version
-  }
+    if (result.values.version) {
+      debug('Inferring app-version from version in package.json')
```
malept: Can we make this debug statement more specific now with `result.source`?

rahatarmanahmed:
Made changes from code review input. Looked into the test coverage, and it looks like the issue is stemming from the reason why @zeke opened #461 in the first place: it's picking up values from `electron-packager`'s own package.json.

@malept's suggestions:

> 1. Copy a fixture to a temporary directory (so that the electron-packager package.json doesn't get detected by get-package-info) - and even then, there could be inconsistent failures on dev machines where someone puts a package.json in their root temporary directory
> 2. Refactor that code to stop at the first package.json - I'm generally in favor of this method, but the concern was that it would break existing code (even though that wasn't an explicit feature)

The issue with option 2 is that the feature of looking at parent package.json's is a required feature for project structures where app code is in a subdirectory with its own package.json. The idea is that the app package.json would have the `productName`, and the parent package.json would have a devDependency on `electron`. Not sure what best practice is nowadays, but I imagine a good number of projects do it this way.

I'm not so sure the issue in option 1 of having a package.json above the temp directory is going to be a huge issue. That seems like it'd be an extremely rare occurrence. If we need to, we can explicitly check if there is a stray package.json and fail the test w/ an informative error msg.

malept:
I'm OK with you implementing option 1 in the way you describe.

zeke:
> looking at parent package.json's is a required feature for project structures where app code is in a subdirectory with its own package.json.

☝️ great point.

> I'm OK with you implementing option 1 in the way you describe.

👍

test/basic.js
```diff
@@ -265,16 +266,52 @@ function createInferElectronTest (opts) {
   }
 }
 
+function copyFixtureToTempDir (fixtureSubdir, cb) {
+  let tmpdir = path.join(os.tmpdir(), fixtureSubdir)
+  let fixtureDir = path.join(__dirname, 'fixtures', fixtureSubdir)
+  waterfall([
+    cb => fs.emptyDir(tmpdir, cb),
+    (cb1, cb2) => fs.copy(fixtureDir, tmpdir, cb2 || cb1), // inconsistent cb arguments from fs.emptyDir
```
rahatarmanahmed: This is weird b/c fs.emptyDir doesn't return a consistent # of arguments, depending on whether the directory already exists or not. I've made an issue on the `fs-extra` repo but this is the workaround for now.

rahatarmanahmed:
love too have tests fail at random on CI

I've run the failing tests on loop on my laptop but I can't figure out what's causing those rare failures. Otherwise, the code is ready to be re-reviewed. Hopefully something pops out during review that tells us about the failures.

eljefedelrodeodeljefe:
I believe landing this is necessary to fix #513

eljefedelrodeodeljefe:
Also highly doubtful that this module should be used at all.

malept:
> Also highly doubtful that this module should be used at all.

Why is that? Do you have an alternative proposal?

eljefedelrodeodeljefe:
So using this is the higher-level of it's upstream dependencies which are already high level of try-catch-ed`JSON.parse(fs.readFile('package.json'))` or `require('package.json')`. Either use its various upstream deps or those of the npm guys or just the above. This would decrease surface area of bugs.

rahatarmanahmed:
If you wanna move the logic from get-package-info into electron-packager, that's fine by me. I initially made it a separate package so it wouldn't muck up the code here, and could be unit tested as well as tested thru electron-packager's integration tests.

(Also I could have sworn the tests were failing before? I don't know how to avoid the coverage decrease, since it's the line that's triggered on some kind of I/O failure looking for package.jsons.

index.js
```diff
@@ -23,50 +23,80 @@ function getMetadata (opts, dir, cb) {
...
   // Search package.json files to infer name and version from
-  getPackageInfo(props, dir, function (err, result) {
-    if (err) {
-      // `get-package-info` exploded looking for `electron`. Try `electron-prebuilt` instead
-      props.pop()
-      props.push(['dependencies.electron-prebuilt', 'devDependencies.electron-prebuilt'])
-      getPackageInfo(props, dir, function (err, result) {
-        if (err) return cb(err)
-        return inferNameAndVersionFromInstalled('electron-prebuilt', opts, result, cb)
+  getPackageInfo(props, dir, (err, result) => {
```
malept: This function has gotten rather large. Can it be split up?

index.js
```diff
@@ -23,50 +23,80 @@ function getMetadata (opts, dir, cb) {
...
+
+        debug(err.message)
+        err.message = messages.join('\n') + '\n'
+        return cb(err)
+      } else {
+        // Missing props not required, can continue w/ partial result
+        result = err.result
+      }
+    } else if (err) {
+      return cb(err)
```
malept: Would an IO error include an invalid `package.json`? Just trying to think of ways to test this line.

index.js
```diff
@@ -23,50 +23,80 @@ function getMetadata (opts, dir, cb) {
...
 
-  if (result.values[`dependencies.${packageName}`]) {
-    resolve(packageName, {
-      basedir: path.dirname(result.source[`dependencies.${packageName}`].src)
-    }, function (err, res, pkg) {
-      if (err) return cb(err)
-      debug(`Inferring target Electron version from ${packageName} dependency or devDependency in package.json`)
-      opts.version = pkg.version
+    if (result.values[`dependencies.electron`]) {
+      var packageName = result.source[`dependencies.electron`].prop.split('.')[1]
```
malept: `let` (I need to add ES6 rules to eslint...)

test/basic.js
```diff
@@ -265,16 +267,57 @@ function createInferElectronTest (opts) {
...
+      packager(opts, function (err, paths) {
+        t.ok(err, 'error thrown')
+        t.end()
+      })
+    })
+  }
+}
+
+function createInferMissingVersionTest (opts) {
+  return function (t) {
```
malept: Should use fat arrow function syntax when possible.

rahatarmanahmed:
@malept moved the electron `package.json` reading into its own function, lmk if you have any better ideas on how to break it up even more.

Good call on the invalid `package.json`. Added a test for that. Should cover that line.

Tests on node v4 seem to be failing due to some eslint peerdependency business, tho... 😞

malept:
> Tests seem to be failing due to some eslint peerdependency business, tho... :disappointed:

WTF. I wanted to add Node 7 to CI anyway, I guess I need to mess with other dependencies too.

malept:
Can't upgrade CI to use Node 7 until next week (blocked on AppVeyor). However, I upgraded `eslint-plugin-promise` on master and CI all passed the `npm install` portion.

index.js
```diff
@@ -23,50 +23,84 @@ function getMetadata (opts, dir, cb) {
...
+    if (err && err.missingProps) {
+      let requiredProps = ['productName', 'dependencies.electron']
+      let missingProps = err.missingProps.map(prop => {
+        return Array.isArray(prop) ? prop[0] : prop
       })
-    } else {
-      return inferNameAndVersionFromInstalled('electron', opts, result, cb)
+
+      // Callback w/ error if there are required props that are missing
+      if (missingProps.filter(prop => requiredProps.find(reqProp => prop === reqProp)).length !== 0) {
```
malept: I'd extract the conditional into a method for readability (and also lets you get rid of the comment.

index.js
```diff
@@ -23,50 +23,84 @@ function getMetadata (opts, dir, cb) {
...
+        let messages = missingProps.map(missingProp => {
```
malept: Perhaps extract the map function into a top-level function named `errorMessageForProperty`?

rahatarmanahmed:
greeeeeeeeen 🤑

zeke:
Thanks for persevering, @rahatarmanahmed 👊

PR #537, opened 2016-12-07
Link to OSX instructions for making a `.icns` file

briandk:
I added a doc link to apple's guide for creating high resolution icons and turning them into an `.icns` file. That way, users who make app icons are better prepared to pass a `--icon` string and get icons on OSX builds.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -124,7 +124,7 @@ The local path to the icon file, if the target platform supports setting embeddi
 
 Currently you must look for conversion tools in order to supply an icon in the format required by the platform:
 
-- OS X: `.icns`
+- [OS X][1]: `.icns` ([Use `iconutil`][2] to convert your PNG icons to a single `.icns` file)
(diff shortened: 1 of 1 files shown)


briandk:
I had forgotten about building a Mac app from other OSes. 

I understand wanting to avoid os-specific instructions, but I think we shouldn't force users to fend for themselves on things we know the answers to. 

For example, on windows OR Mac one can use this: https://iconverticons.com/online/

Free for online use, $5 to get the app for Windows/Linux

malept:
I'm not saying I reject your changes. I'm saying that if you're going to have OS-specific instructions, we need to give "equal time" to all of the host platforms that we support. Preferably a free (non-web), open source alternative on Windows.

malept:
I was thinking about it this morning - this might be better off either in the Electron application distribution guide, or Electron Packager needs a guide of its own.

malept:
I'm closing this since I think this sort of documentation should really go into the Electron docs (and it should cover all major OSes, not just macOS). I've seen people ask about `.icns` files not in the context of Electron Packager, in the Electron community forums. I will be happy to link to the Electron docs that cover this, though.

PR #540, opened 2016-12-08
Fix sign task to propagate error back up

emorikawa:
When the codesign fails due to an error like `Error: No identity found for signing`, not propagating the error will not notify the remaining dependent tasks. The builds were being marked as done with no errors, even when codesign failed.

### DIFF
--- a/mac.js
+++ b/mac.js
@@ -212,7 +212,7 @@ class MacApp {
             // Although not signed successfully, the application is packed.
             console.warn('Code sign failed; please retry manually.', err)
           }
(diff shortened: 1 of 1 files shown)


malept:
Could you add a regression test for this?

malept:
Since the original issue reporter has not responded to the request for changes in about a month, I'm going to assume that they either don't have the time to work on this or decided not to pursue this PR. If I'm incorrect, they can respond with an updated pull request, and I'll reopen.

PR #546, opened 2016-12-22
add wine download path

CooperFu:
just add wine download path , because when I use `homebrew` they tell me download form this webside.


malept:
What exactly is the message when you use homebrew? Because I don't want to link to a specific Wine version announcement unless it's necessary.

PR #568, opened 2017-01-27
Updated "OS X" to "macOS"

amitmerchant1990:
Hi team,

I've updated the reference of "OS X" to "macOS" to keep up with the new name given to OS X by Apple.

Thanks!


malept:
Thanks for the pull request, but I'm opting to just modify just one line in the readme: fe5b35b7145710fc16ef00c3362ff0927adf6771

PR #580, opened 2017-02-07
Use camelCase identifiers in the JS API

### LINKED ISSUE #325: Programmatic API options should follow Javascript naming conventions

j-f1:
- [x] I have read the [section in CONTRIBUTING.md about pull requests](https://github.com/electron-userland/electron-packager/blob/master/CONTRIBUTING.md#filing-pull-requests).
- [x] My changes are appropriately documented.
- [ ] My changes have sufficient test coverage.
  - Does it need any tests?
- [x] The test suite passes.

This fixes #325.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### Changed
+* Options previously formatted in kebab-case (i.e., with hyphens) are now available in camelCase, per JavaScript naming standards (#580)
+
+### Deprecated
+* Options currently formatted in kebab-case (i.e., with hyphens) are deprecated in favor of their camelCase variants, per JavaScript naming standards (#580)
+
 ## [8.5.2] - 2017-02-19
 
 ### Fixed
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,10 +3,12 @@
 ## Unreleased
 
 ### Changed
-* Options previously formatted in kebab-case (i.e., with hyphens) are now available in camelCase, per JavaScript naming standards (#580)
+
+* Options formatted in kebab-case (i.e., with hyphens) are available in camelCase, per JavaScript naming standards (#580)
 
 ### Deprecated
-* Options currently formatted in kebab-case (i.e., with hyphens) are deprecated in favor of their camelCase variants, per JavaScript naming standards (#580)
+
+* Options formatted in kebab-case (i.e., with hyphens) are deprecated in favor of their camelCase variants, per JavaScript naming standards (#580)
 
 ## [8.5.2] - 2017-02-19
 
--- a/common.js
+++ b/common.js
@@ -40,8 +40,11 @@ function parseCLIArgs (argv) {
   args.dir = args._[0]
   args.name = args._[1]
 
+  // Transform hyphenated keys into camelCase
+  module.exports.camelCase(args, false)
+
   var protocolSchemes = [].concat(args.protocol || [])
-  var protocolNames = [].concat(args['protocol-name'] || [])
+  var protocolNames = [].concat(args.protocolName || [])
 
   if (protocolSchemes && protocolNames && protocolNames.length === protocolSchemes.length) {
     args.protocols = protocolSchemes.map(function (scheme, i) {
@@ -53,10 +56,6 @@ function parseCLIArgs (argv) {
     args.out = null
   }
 
-  // Transform hyphenated keys into camelCase
-  args.electronVersion = args['electron-version']
-  delete args['electron-version']
-
   // Overrides for multi-typed arguments, because minimist doesn't support it
 
   // asar: `Object` or `true`
@@ -65,8 +64,8 @@ function parseCLIArgs (argv) {
   }
 
   // osx-sign: `Object` or `true`
-  if (args['osx-sign'] === 'true') {
-    args['osx-sign'] = true
+  if (args.osxSign === 'true') {
+    args.osxSign = true
   }
 
   // tmpdir: `String` or `false`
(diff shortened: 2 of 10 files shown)


common.js
```diff
@@ -188,6 +187,34 @@ module.exports = {
...
+    'osx-sign': 'osxSign',
+    'protocol-name': 'protocolName',
+    'version-string': 'versionString'
+  },
+
+  camelCase: function camelCase (properties, warn) {
+    Object.keys(module.exports.kebabProperties).forEach(function (key) {
+      var value = module.exports.kebabProperties[key]
+      if (properties.hasOwnProperty(key)) {
+        warn && warning(`The ${key} parameter is deprecated when used from JS, use ${value} instead`)
```
malept: I would prefer it if this was a full-fledged `if` statement.

malept: Also, it should be mentioned that the old property is going to be removed in the next major version.

malept:
@j-f1 It looks like there's a problem in `win32.js`, take a look at the test output.

Also, what part of the testsuite is not running locally?

malept:
Could you also add a NEWS entry? (An entry under both "Changed" and "Deprecated".)

j-f1:
@malept  Sorry for the delay. I’ve added the changes requested.

The problem doesn’t make sense. Presumably, the files are minified somehow (see [this line](https://travis-ci.org/electron-userland/electron-packager/jobs/199956707#L4329)).

When I try to run the test locally, it downloads Electron, then nothing happens.

NEWS.md
```diff
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### Changed
+* The JS API now uses camelCase identifiers.
```
malept: > Options previously formatted in kebab-case (i.e., with hyphens) are now available in camelCase, per JavaScript naming standards (#580)

NEWS.md
```diff
@@ -2,6 +2,12 @@
...
+
+### Deprecated
+* Using kebab-case keys for the JS API is deprecated. Please use camelCase.
```
malept: Options currently formatted in kebab-case (i.e., with hyphens) are deprecated in favor of their camelCase variants, per JavaScript naming standards (#580)

docs/api.md
```diff
@@ -51,13 +51,13 @@ An array of functions to be called after Electron has been extracted to a tempor
 
 When `true`, sets both [`arch`](#arch) and [`platform`](#platform) to `all`.
 
-##### `app-copyright`
+##### `appCopyright`
```
malept: For the docs, what I typically do when I'm renaming an option is add an entry for the old name with the type and a deprecated message (for example: [`version-string`](https://github.com/electron-userland/electron-packager/blob/v8.5.1/docs/api.md#version-string))

common.js
```diff
@@ -188,6 +187,36 @@ module.exports = {
...
+    'app-version': 'appVersion',
+    'build-version': 'buildVersion',
+    'app-bundle-id': 'appBundleId',
+    'app-category-type': 'appCategoryType',
+    'extend-info': 'extendInfo',
+    'extra-resource': 'extraResource',
+    'helper-bundle-id': 'helperBundleId',
+    'osx-sign': 'osxSign',
+    'protocol-name': 'protocolName',
+    'version-string': 'versionString'
```
malept: There's no point in camel-casing `version-string`, it's deprecated.

malept:
> The problem doesn’t make sense. Presumably, the files are minified somehow

I think this is an artifact of the code coverage module.

win32.js
```diff
@@ -29,23 +29,23 @@ module.exports = {
...
-      if (opts['build-version']) {
-        rcOpts['file-version'] = opts['build-version']
+      if (opts.buildVersion) {
+        rcOpts['file-version'] = opts.buildVersion
       }
 
-      if (opts['app-copyright']) {
-        rcOpts['version-string'].LegalCopyright = opts['app-copyright']
+      if (opts.appCopyright) {
+        rcOpts.versionString.LegalCopyright = opts.appCopyright
```
malept: This is incorrect, `rcOpts` is for the `rcedit` module. I take it you did a global search/replace?

win32.js
```diff
@@ -29,23 +29,23 @@ module.exports = {
...
       }
 
-      if (opts.icon || opts.win32metadata || opts['version-string'] || opts['app-copyright'] || opts['app-version'] || opts['build-version']) {
+      if (opts.icon || opts.win32metadata || opts.versionString || opts.appCopyright || opts.appVersion || opts.buildVersion) {
```
malept: Revert the `opts.versionString`, please.

docs/api.md
j-f1: @malept Should I place the deprecated names right underneath the new one, or should I group them at the bottom?

malept: Please keep the options alphabetized, so probably place the deprecated options somewhere above their respective new names.

docs/api.md
```diff
@@ -247,38 +279,58 @@ Valid values are listed in [Apple's documentation](https://developer.apple.com/l
...
+
 *Object* or *`true`*
 
 If present, signs OS X target apps when the host platform is OS X and XCode is installed. When the value is `true`, pass default configuration to the signing module. The configuration values listed below can be customized when the value is an `Object`. See [electron-osx-sign](https://www.npmjs.com/package/electron-osx-sign#opts) for more detailed option descriptions and the defaults.
 - `identity` (*String*): The identity used when signing the package via `codesign`.
 - `entitlements` (*String*): The path to the 'parent' entitlements.
 - `entitlements-inherit` (*String*): The path to the 'child' entitlements.
 
+
+
```
malept: Unnecessary whitespace

docs/api.md
```diff
@@ -116,6 +132,12 @@ but are not limited to:
 - `strictSSL` (*Boolean* - default: `true`): Whether SSL certificates are required to be valid when
   downloading Electron.
 
+
+##### `electron-version`
+
+*String* (**deprecated** and will be removed in a future major version, please use the
+[`electronVersion`](#electronVersion) parameter instead)
+
```
malept: This never existed as a JavaScript API option.

PR #605, opened 2017-03-02
Always return an Error or null in node-style callbacks

feross:
This will be a style error in the next version of standard (v10).


malept:
Thanks for the PR! The test needs fixing, too.

feross:
Fixed tests.

PR #610, opened 2017-03-06
Add support for updating the application manifest and requested execution level for Windows targets

### LINKED ISSUE #197: Add support for setting a Windows manifest different from the embedded Electron default
I used the winreg npm package for my app and I need elevated privileges for the uses I have.

I would like to associate a manifest file with an app build with electron (atom-shell) and packaged on windows with electron-packager but I can not find any way to do it easily.

An old app of mine was made with python and cx_freeze and it sufficed to put the file with the name app.exe.manifest in the same directory.

I can't understand why it doesn't work this time.

dmascord:

**Summarize your changes:**

As per http://layer0.authentise.com/electron-and-uac-on-windows.html, have added the ability for rcedit.exe to modify the requested execution level, so that the Electron app can have administrator privileges.  https://github.com/dmascord/rcedit and https://github.com/dmascord/node-rcedit have been updated appropriately, and a pull request has been issued for them too.

Fixes #197.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "get-package-info": "^1.0.0",
     "minimist": "^1.1.1",
     "plist": "^2.0.0",
-    "rcedit": "^0.8.0",
+    "rcedit": "^0.9.0",
     "resolve": "^1.1.6",
     "run-series": "^1.1.1",
     "sanitize-filename": "^1.6.0",
--- a/test/win32.js
+++ b/test/win32.js
@@ -93,15 +93,30 @@ function setCopyrightAndCompanyNameTest (appCopyright, companyName) {
 function setRequestedExecutionLevelTest (requestedExecutionLevel) {
   var opts = {
     win32metadata: {
-      requestedExecutionLevel: 'asInvoker'
+      'requested-execution-level': requestedExecutionLevel
     }
   }
 
   return generateVersionStringTest(
-    'win32metadata',
+    'requested-execution-level',
     opts,
-    {requestedExecutionLevel: requestedExecutionLevel},
-    'requestedExecutionLevel in win32metadata should match win32metadata value '
+    requestedExecutionLevel,
+    'requested-execution-level in win32metadata should match rcOpts value'
+  )
+}
+
+function setApplicationManifestTest (applicationManifest) {
+  var opts = {
+    win32metadata: {
+      'application-manifest': applicationManifest
+    }
+  }
+
+  return generateVersionStringTest(
+    'application-manifest',
+    opts,
+    applicationManifest,
+    'application-manifest in win32metadata should match rcOpts value'
   )
 }
 
@@ -174,4 +189,5 @@ util.packagerTest('win32 app copyright sets LegalCopyright test', setCopyrightTe
 util.packagerTest('win32 set LegalCopyright and CompanyName test', setCopyrightAndCompanyNameTest('Copyright Bar', 'MyCompany LLC'))
 util.packagerTest('win32 set CompanyName test (win32metadata)', setCompanyNameTest('MyCompany LLC', 'win32metadata'))
 util.packagerTest('win32 set CompanyName test (version-string)', setCompanyNameTest('MyCompany LLC', 'version-string'))
-util.packagerTest('win32 set requestedExecutionLevel test (win32metadata)', setRequestedExecutionLevelTest('asInvoker'))
+util.packagerTest('win32 set requested-execution-level test (win32metadata)', setRequestedExecutionLevelTest('asInvoker'))
+util.packagerTest('win32 set application-manifest test (win32metadata)', setApplicationManifestTest('/path/to/manifest.xml'))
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "electron-packager",
-  "version": "8.5.2",
+  "version": "8.5.3",
```
MarshallOfSound: Don't bump this version yourself 👍

win32.js
```diff
@@ -44,8 +44,14 @@ module.exports = {
       if (opts.appCopyright) {
         rcOpts['version-string'].LegalCopyright = opts.appCopyright
       }
+      
+      // add "asInvoker" to the built manifest by default
+      rcOpts['requested-executation-level'] = 'asInvoker'
+      if (opts.win32exeLevel) {
```
MarshallOfSound: This option should probably go inside `win32metadata` IMO

malept: I agree with @MarshallOfSound.

dmascord: The manifest (http://msdn2.microsoft.com/en-us/library/1w45z383.aspx) changes are not metadata (http://msdn2.microsoft.com/en-us/library/ms404430.aspx) changes, so it would be a bit counter-intuitive to overload that option.

Would a new option, win32manifest.requestedExecutionLevel or similar be more appropriate here ?

package.json
malept: This wouldn't be a patch level change anyway.

win32.js
malept: That depends: what other manifest options might be useful to set in the future (via future rcedit features)?

dmascord: Potentially we would want to be able to Strong Name the assembly: https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/strong-name-signing.md
There are also a few settings, such as dpiAware, and supportedOS that we may want to insert into the manifest.

malept: Electron apps on Windows have a .NET assembly?

Those other settings sound like things that Electron should already set, but supportedOS might be useful if the app depends on another exe with a higher Windows requirement...

malept: If we're going to have a new option, I'd want to have tests for it.

dmascord: How do you want the tests handled?  The tests that I've seen only really test if the options actually get set, rather than test that the option setting did anything.    There are functional tests in node-rcedit now, are these adequate or do you need something else ?

win32.js
```diff
@@ -44,8 +44,14 @@ module.exports = {
       if (opts.appCopyright) {
         rcOpts['version-string'].LegalCopyright = opts.appCopyright
       }
+      
+      // add "asInvoker" to the built manifest by default
+      rcOpts['requested-executation-level'] = 'asInvoker'
```
malept: Misspelled again :smile:

win32.js
malept: > The tests that I've seen only really test if the options actually get set, rather than test that the option setting did anything.

Yeah, I'd prefer these tests. There are too many integration tests as it is, which is why the testsuite takes so long to run. I just want tests that assert that the default and custom values get set appropriately.

win32.js
dmascord: gah! My brain tells my fingers the wrong thing somehow! :)  Will fix now

win32.js
malept: I've talked to @MarshallOfSound about this, we've come to the conclusion that it should go into `win32metadata`, despite it being not quite metadata. I'd like the property to be called `requestedExecutionLevel` since that's the name of the manifest property (according to that blog post. _(The other properties are the case-sensitive property names set by rcedit.)_

malept:
Hey, I've merged #612. The downside is that you need to resolve a merge conflict. The upside is that it should be easier to write unit tests for your feature.

malept:
Do you mind rebasing? GitHub's diff has gotten rather messy.

dmascord:
No worries, will write the test first, and then rebase

test/win32.js
```diff
@@ -90,6 +90,21 @@ function setCopyrightAndCompanyNameTest (appCopyright, companyName) {
   )
 }
 
+function setRequestedExecutionLevelTest (requestedExecutionLevel) {
+  var opts = {
+    win32metadata: {
+      requestedExecutionLevel: 'asInvoker'
```
malept: Isn't this the default?

dmascord: I've decided to remove the default, and by default don't modify the assembly.   Whatever upstream wants to put as the default in the electron.exe should probably stay as the default.

malept: Sounds reasonable.

dmascord:
Rebased my changes (sorry for the delay)

malept:
Hmm. Your rebase caused the GitHub diff to be really weird. In any event it's probably not worth rebasing until the other two PRs are dealt with.

dmascord:
Maybe create a new pull ?

debelop13:
How it goes? Thanks!

malept:
Still blocked on the other two PRs.

malept:
I'm going to rebase this PR this weekend and see what's left, because I've forgotten. The PRs that were blocking this are now merged/released. From CI, it looks like there's at least linting failures. And the docs still need to be updated.

Issue #518, opened 2016-10-28
Allow yarn to be used instead of npm


soredake:
Lines like this preventing normal use with yarn:
`child.exec('npm prune --production', {cwd: appPath}, cb)`
`exec('npm install --no-bin-links'`

malept:
That's preventing yarn completely replacing npm, sure. In the meantime you should be able to symlink `npm` to `yarn` and in theory it should "just work".

soredake:
yarn prune has no --production option
yarn install has no --no-bin-links option, so just symlink is not the solution (while sed -i /npm --option/yarn --option, but it is not so convenient)

malept:
Then it sounds like yarn support specifically is blocked on `prune --production` being added. That is an important part of how we package Electron apps.

soredake:
https://github.com/yarnpkg/yarn/issues/696#issuecomment-253031239

malept:
The above yarn issue relates to `prune`. This one relates to `--no-bin-links`: https://github.com/yarnpkg/yarn/issues/929

Those two issues block proper yarn support. However, I have spec'd out how to implement generic NPM-compatible package manager support in https://github.com/electron-userland/electron-packager/issues/515#issuecomment-257167905, and I am happy to review pull requests that implement this (and, obviously, follow the [contributing guidelines](https://github.com/electron-userland/electron-packager/blob/master/CONTRIBUTING.md#filing-pull-requests)).

soredake:
https://github.com/yarnpkg/yarn/pull/1651

soredake:
https://github.com/yarnpkg/yarn/issues/696#issuecomment-258418656

jrobeson:
`--no-bin-links` did finally happen, but still no resolution for the other issue.

Issue #515, opened 2016-10-28
Allow cnpm to be used instead of npm on Windows


TooBug:
cnpm is a replacement tool of npm which optimized the install speed. BTW, it uses a Chinese mirror by default.

The structure of node_modules is quite different when modules are installed with cnpm. (It uses some soft links.)

The installation of cnpm is:

```
npm install -g cnpm
```

When package apps, electron-packager hangs. The log stopped by this: 

```
electron-packager Ignored paths based on the out param: +3ms [ '/Users/TooBug/app-darwin-ia32',
  '/Users/TooBug/app-darwin-x64',
  '/Users/TooBug/app-darwin-armv7l',
  '/Users/TooBug/app-linux-ia32',
  '/Users/TooBug/app-linux-x64',
  '/Users/TooBug/app-linux-armv7l',
  '/Users/TooBug/app-mas-ia32',
  '/Users/TooBug/app-mas-x64',
  '/Users/TooBug/app-mas-armv7l',
  '/Users/TooBug/app-win32-ia32',
  '/Users/TooBug/app-win32-x64',
  '/Users/TooBug/app-win32-armv7l' ]
```

And the template folder of app is keeping increase, never stops.

I guess the problem occurs on the `npm prune` step.

related issue:  https://github.com/cnpm/npminstall/issues/155

malept:
Does `cnpm prune --production` exist? If so, if  you modify Electron Packager to use that command instead of `npm prune --production`, does it still hang?

TooBug:
Yes, cnpm do has the same `prune` command. How can I force electron-packager to use that command?

malept:
Electron Packager needs to be changed to be able to specify a custom command as the package manager, in order to run the "prune `devDependencies`" command. (And also an `npm install` command in the testsuite.)

It should be called `packageManager` in the API, and `--package-manager` in the CLI. The default value should be `npm`. It should fail with a nice error message if the specified package manager either does not exist or is not an executable file.

malept:
I am having a really difficult time getting cnpm to be called correctly on Windows, for some reason. Since Windows is not my area of expertise, I'm inclined to only implement cnpm support for Linux/macOS and wait for someone to file a PR to get this working correctly on Windows. It will raise an error if cnpm is selected on Windows in the meantime.

PR #614, opened 2017-03-08
Add max-buffer-size option for the parameter for pruning child process

hrkt:

**Summarize your changes:**

Hi.

I think it might be better if this command has "max-buffer-size" cli option (optional).

This patch enables users to avoid "stdout maxBuffer exceeded" error when
the size of pruning targets is larger than 200MB (default for child.exec).

In my case, the existence of many dependencies in package.json causes  "stdout maxBuffer exceeded" without "--no-prune" option.

### DIFF
--- a/common.js
+++ b/common.js
@@ -33,6 +33,7 @@ function parseCLIArgs (argv) {
     },
     string: [
       'electron-version',
+      'max-buffer-size',
       'out'
     ]
   })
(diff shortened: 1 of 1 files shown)


malept:
Thanks for the pull request. I'm worried that this is a pretty small edge case that will make using `prune` (and Electron Packager) more confusing.

Instead, could you try to install `yarn` and replace `npm prune --production` with `yarn install --production` in the code? I want to see if using an alternate package manager solves this problem for you.

hrkt:
Thank you for kind advice.

I tried editing common.js.

While
        // child.exec('npm prune --production', {cwd: appPath}, cb)
fails in my case (app size > 400MB), 
        child.exec('yarn install --production', {cwd: appPath}, cb)
works like a charm.

malept:
I would rather have a pull request that implements https://github.com/electron-userland/electron-packager/issues/518 (and https://github.com/electron-userland/electron-packager/issues/515 - the spec for both is in https://github.com/electron-userland/electron-packager/issues/515#issuecomment-257167905). Are you interested?

hrkt:
I read the #515, #518. The proposed idea of the packager option seems useful.

I'll try making new PR in this weekend.

PR #618, opened 2017-03-12
Add package-manager option

### LINKED ISSUE #518: Allow yarn to be used instead of npm

hrkt:
* [?] The changes are appropriately documented (if applicable).
* [?] The changes have sufficient test coverage (if applicable).

**Summarize your changes:**

This PR introduces "package-manager" option.
It is a replacement of #614, and it tries to implement the spec defined in #515.

It can parse 'nom', 'yarn' as package-manager option.
When other package manager is specified, a warning message is shown.

This PR does not include the code below.
1. modification of usage.txt 
2. unit test

Thanks in advance for reading this message.

Fixes #518.
Addresses #515.

### DIFF
--- a/common.js
+++ b/common.js
@@ -88,6 +88,18 @@ function asarApp (appPath, asarOptions, cb) {
   })
 }
 
+function pruneModules (opts, appPath, cb) {
+  var pmName = opts['package-manager']
+  if (pmName !== undefined && pmName === 'yarn') {
+    child.exec('yarn install --production ', { cwd: appPath }, cb)
+  } else if (pmName !== undefined && pmName !== 'npm') {
+    warning(`specified package-manager "${pmName}" is not available, use {"npm", "yarn"} instead`)
+  } else {
+    // defaults to npm
+    child.exec('npm prune --production', { cwd: appPath }, cb)
+  }
+}
+
 function isPlatformMac (platform) {
   return platform === 'darwin' || platform === 'mas'
 }
@@ -293,8 +305,7 @@ module.exports = {
     // appPath is predictable (e.g. before .app is renamed for mac)
     if (opts.prune || opts.prune === undefined) {
       operations.push(function (cb) {
-        debug('Running npm prune --production')
-        child.exec('npm prune --production', {cwd: appPath}, cb)
+        pruneModules(opts, appPath, cb)
       })
     }
 
--- a/common.js
+++ b/common.js
@@ -1,7 +1,6 @@
 'use strict'
 
 const asar = require('asar')
-const child = require('child_process')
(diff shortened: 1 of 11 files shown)


common.js
```diff
@@ -88,6 +88,18 @@ function asarApp (appPath, asarOptions, cb) {
   })
 }
 
+function pruneModules (opts, appPath, cb) {
+  var pmName = opts['package-manager']
```
malept: per the spec:

> It should be called packageManager in the API

common.js
```diff
@@ -88,6 +88,18 @@ function asarApp (appPath, asarOptions, cb) {
...
+  if (pmName !== undefined && pmName === 'yarn') {
+    child.exec('yarn install --production ', { cwd: appPath }, cb)
+  } else if (pmName !== undefined && pmName !== 'npm') {
+    warning(`specified package-manager "${pmName}" is not available, use {"npm", "yarn"} instead`)
+  } else {
+    // defaults to npm
+    child.exec('npm prune --production', { cwd: appPath }, cb)
+  }
```
malept: There's a better way to structure this:

```javascript
opts.packageManager ||= 'npm'

switch (packageManager) {
  case 'npm':
    // do stuff
  case 'yarn':
    // do stuff
  default:
    throw error;
}
```

common.js
```diff
@@ -293,8 +305,7 @@ module.exports = {
     // appPath is predictable (e.g. before .app is renamed for mac)
     if (opts.prune || opts.prune === undefined) {
       operations.push(function (cb) {
-        debug('Running npm prune --production')
```
malept: This should come back in some form or another. We need to be able to see when the prune operation occurs and with what package manager. It's pretty helpful for debugging user problems.

hrkt:
Thank you, @malept for the review. I'll fix it in this week.

malept:
For tests, I think what I'd like to do is have Travis/AppVeyor download Yarn/cnpm and make sure they work correctly with pruning.

The test reference to NPM should probably be configurable via an environment variable.

malept:
I've refactored this PR so that it also works with `cnpm`, and added docs/tests.

> The test reference to NPM should probably be configurable via an environment variable.

I decided against this. I'm OK with having people working on tests needing to have NPM installed.

malept:
Need to fix calling cnpm on Windows (per CI failure) and also add a NEWS entry.

malept:
I'm no longer calling cnpm on Windows (see: https://github.com/electron-userland/electron-packager/issues/515#issuecomment-297604044).

PR #658, opened 2017-05-29
Promisify tests and packager function

### LINKED ISSUE #627: return promise
so we can use async/await instead of callback

malept:

**Summarize your changes:**

Converts all tests to use Promises (and gets rid of the waterfall dependency), and adds Promise support to the `packager` function via `pify`.

This is technically a breaking change, because the return value of `packager` has changed.

Fixes #627.


Jmclerck:
I assume from looking at the release notes for the most recent versions of electron-packager this change hasn't been released yet? If not the Docs are a little confusing as it suggests that the Promisified versions of the API should already work when in fact they don't.

Is there an ETA on a release containing these changes?

Thanks!

malept:
All of the docs are based on the current state of `master`. If you want the docs for the released version, you'll need to look at the docs for a given version's tag.

I have about 3 more breaking changes to do before I release, and I do all of this work in my spare time, so I don't have an ETA, sorry.

Jmclerck:
Hi @malept thanks for the clarification and quick reply. No problem about the ETA, I just wanted to check in case you had any idea.

Thanks for all your work on this, it is appreciated!

PR #775, opened 2017-12-12
Moved hostArch() earlier in validateListFromOptions()

mathuin:

**Summarize your changes:**

This PR is in response to atom/atom#15431.  When I tried to build Atom on my Chromebook running Arch Linux ARM, the automatic trigger detecting armv7l was not running because a value already existed for the arch.  This change moves the hostArch() function before the list assignment to catch this case.

I did not add documentation for this change because it is sufficiently simple that I think additional comments would not be helpful.  Similarly, I did not add tests.  I confess I don't know how to run the test suite, but the code did pass one important test -- when I ran it, the correct architecture was detected and
a package was built!

### DIFF
--- a/targets.js
+++ b/targets.js
@@ -102,13 +102,14 @@ module.exports = {
   validateListFromOptions: function validateListFromOptions (opts, name) {
     if (opts.all) return Array.from(supported[name].values())
 
-    let list = opts[name]
+    let list
+    if (name === 'arch') {
+      list = hostArch()
+    } else {
+      list = opts[name]
+    }
     if (!list) {
-      if (name === 'arch') {
-        list = hostArch()
-      } else {
-        list = process[name]
-      }
+      list = process[name]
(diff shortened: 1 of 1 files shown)


malept:
The big problem with this solution is that the tests fail. Frankly, I'm not sure it's the correct solution.

mathuin:
That's fine -- the intent was to share with you the code that ran on my system in the hopes that you would be able to make a proper patch.  Like I mentioned in the other issue, I don't know the language and certainly don't understand the codebase.  It's okay if you want to close the PR -- I just wanted you to see the code that worked here.

malept:
I'm going to close this due to the rationale I laid out in https://github.com/atom/atom/pull/15431#issuecomment-353712740 - in the near future, I'll add explicit unit tests for `hostArch()` and `validateListFromOptions()` regarding `armv7l` detection.

PR #790, opened 2018-01-15
Add extraResource custom subpath support

Hammster:
* [ ] The changes have sufficient test coverage (if applicable).
* [x] The test suite passes successfully on my local machine (if applicable).

**Summarize your changes:**

Added a configuration object for ExtraResource, it allows to define a destination inside the resources folder or to rename the file. This is also in the feature set of `electron-builder` and is applicable to use cases where e.g you need to pick one certain DLL from within devDependencies to your resource folder.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -153,10 +153,14 @@ valid versions. If omitted, it will use the version of the nearest local install
 
 ##### `extraResource`
 
-*String* or *Array* of *String*s
+*Object*, *String* or *Array* of *String*s and/or *Object*s
 
 One or more files to be copied directly into the app's `Contents/Resources` directory for OS X
 target platforms, and the `resources` directory for other target platforms.
+
+A Object defines the files source with a custom destination, by the following properties
+- `from` (*String*): The source file
+- `to` (*String*): Destination path for the new file inlclusive the filename, relative from the `resources` directory.
 
 ##### `executableName`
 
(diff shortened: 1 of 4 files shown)


malept:
> is applicable to use cases where e.g you need to pick one certain DLL from within devDependencies to your resource folder.

I'm not understanding why this is a use case. Normally you'd want to prune `devDependencies`, and even if you don't for some reason, you'd want to put the DLL in the same folder as the Electron binary. In that case, you'd want to use the `afterExtract` hook to copy the file into that folder, instead of using `extraResource` (which copies it into the resources folder).

Hammster:
@malept we use this to select the electron version specific DLL of `electron-edge-js` on mac builds previously we used `electron-builder` where the option extraResources / extraFiles allow for individual file parsing into content and resources.

But maybe the use case is not well chosen, let's say you want to keep a folder structure when copying resources then this option will help you archiving that.

malept:
I think this functionality would be better suited for an Electron Packager plugin. I would prefer to avoid making `extraResource` more complicated to understand.

To make into a plugin, you would write an [`afterCopy` hook](https://github.com/electron-userland/electron-packager/blob/v10.1.2/docs/api.md#aftercopy) to derive the resources path from the `buildPath`, and then copy your files to that folder as you see fit.

PR #817, opened 2018-03-24
Add link to electron-installer-windows

fcastilloec:

**Summarize your changes:**
Add a link to `electron-installer-windows` project to the README

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -186,7 +186,7 @@ Windows:
 * [electron-winstaller](https://github.com/electron/windows-installer) - Squirrel.Windows-based installer
 * [electron-windows-store](https://github.com/felixrieseberg/electron-windows-store) - creates an AppX package for the Windows Store
 * [electron-wix-msi](https://github.com/felixrieseberg/electron-wix-msi) - creates traditional MSI installers
-* [electron-installer-windows](https://github.com/unindented/electron-installer-windows) - Squirrel.Windows-based installer
+* [electron-installer-windows](https://github.com/unindented/electron-installer-windows) - alternative Squirrel.Windows-based installer
 
 OS X:
 
--- a/readme.md
+++ b/readme.md
@@ -183,7 +183,7 @@ The `Foo Bar.app` folder generated can be executed by a system running OS X, whi
 
 Windows:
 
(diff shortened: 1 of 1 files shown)


MarshallOfSound:
I'm somewhat confused as to the difference between this and the officially maintained (and already linked) `electron-winstaller`?  Can you explain the differences?

fcastilloec:
`electron-winstaller` is only partially maintained, they only update Squirrel once in a while. There are no test suites, they are not using the latest version of Nuget and it doesn't work on Linux/OSX systems using `mono`/`wine`.
Also, in a project I was contributing, `electron-winstaller` installers were creating errors on installation (see https://github.com/headsetapp/headset-electron/issues/129) which I could replicate with other projects.

On the other hand, `electron-installer-windows` has tests, works in all OSs without any problems (signing the installer is only supported on Windows), it has a CLI, and it has a familiar form, similar to the already used `electron-installer-debian`, `electron-installer-redhat`, etc.

I think it's better to provide users more alternatives, especially if the new alternative works better and in all OSs.
I understand if there's a motivation to only keep one option, or to advertise a specific module, feel free to close this PR.

fcastilloec:
`electron-winstaller` doesn't work for me and a lot of other people according to the issues' page. This is especially true if using Linux or OSX, even all testing for Travis-CI has been removed.
I completely understand why not trusting a project that includes compiled binaries, you might not know where they come from. I'm familiar with `electron-installer-windows` so I'm probably going to push a PR there that downloads the `.zip` directly from Github releases when installing with `npm`. That way users are always sure where things come from. You should consider doing the same for `electron-winstaller`.

I love contributing to projects, but sometimes I need things to "just work". I had so much trouble with `electorn-winstaller`, I actually did try to debug the problem but wasn't getting anywhere as you can see on the link I previously posted. It was taking too much of my time so I switched to `electron-installer-windows`. I would love to try to fix the issues but it will require a lot of time that currently, I don't have. Perhaps in the future.

I think I'll rest my case and like I mentioned originally, feel free to close this PR if you think it's better that way. I really don't mind.

malept:
I'm just going to take this. I made a few changes that differentiates the two links.

Issue #161, opened 2015-10-16
Specify already packed asar via a prebuiltAsar option


kurisubrooks:
I have to do a bunch of things with asar manually in order to get my build to work correctly.
Is there a way you could add a CLI implementation to specify an already packed asar?

malept:
Seems reasonable.

maxogden:
Here's the current asar flag in the code https://github.com/maxogden/electron-packager/blob/437bb530570186a6141c55b2a967615180c5259a/common.js#L94

We could try overloading it. It's currently a boolean, so if it was a string we could try and read an asar file at the path in the string. if it's a boolean we would do what we do now. This might introduce some bugs with CLI parsing so we would have to be sure our CLI parser always turns `--asar=true` into `true` and not `'true'`.

The other option is to add another flag, e.g. `--asar-file=foo.asar`

kurisubrooks:
The option to specify the file/file's directory would be very useful.

malept:
Given the current state of the asar code, adding a new sub-property to `asar` would be best.

if `asar.filename` is specified, it should bypass [the asar creation code](https://github.com/electron-userland/electron-packager/blob/0ef914f/common.js#L239-L241) and just copy the specified asar file to [`app.asar`](https://github.com/electron-userland/electron-packager/blob/0ef914f/common.js#L75).

As always, please make sure that [there are appropriate docs and tests added](https://github.com/electron-userland/electron-packager/blob/master/CONTRIBUTING.md#filing-pull-requests).

jsg2021:
any movement on this? :)

malept:
@jsg2021 I would be more than happy to review a PR to add this feature. If you need any more guidance than what I described in https://github.com/electron-userland/electron-packager/issues/161#issuecomment-252740910 (the code has changed a lot since then), please let me know.

PR #823, opened 2018-04-10
Add support for prebuilt asars

jsg2021:

**Summarize your changes:**
I implemented the --asar.filename option referenced in #161


.editorconfig
```diff
@@ -0,0 +1,17 @@
...
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.md]
+trim_trailing_whitespace = false
+
+[package.json]
+indent_style = space
+indent_size = 2
```
malept: If we're going to add this, it should be in a separate PR.

NEWS.md
```diff
@@ -4,6 +4,8 @@
 
 [Unreleased]: https://github.com/electron-userland/electron-packager/compare/v12.0.0...master
 
+* Added `asar.filename` to support pre-built asar files.
```
malept: Could you please adjust this to conform to the [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) format?

docs/api.md
```diff
@@ -183,12 +183,14 @@ is not restricted to the official set if [`download.mirror`](#download) is set.
...
+- `filename` (*String*): circumvents the asar generation and lets you specify your own pre-built asar file/directory.
 - `ordering` (*String*): A path to an ordering file for packing files. An explanation can be found on the [Atom issue tracker](https://github.com/atom/atom/issues/10163).
 - `unpack` (*String*): A [glob expression](https://github.com/isaacs/minimatch#features), when specified, unpacks the file with matching names to the `app.asar.unpacked` directory.
 - `unpackDir` (*String*): Unpacks the dir to the `app.asar.unpacked` directory whose names exactly or pattern match this string. The `asar.unpackDir` is relative to `dir`.
 
   Some examples:
 
+  - `asar.filename = './dist/app.asar'` will short-circuit creating the asar, and just copy that one into the electron bundle.
```
malept: I don't think there needs to be an example for this, it should be self-explanatory.

docs/api.md
```diff
@@ -183,12 +183,14 @@ is not restricted to the official set if [`download.mirror`](#download) is set.
 *Boolean* or *Object* (default: `false`)
 
 Whether to package the application's source code into an archive, using [Electron's archive format](https://github.com/electron/asar). Reasons why you may want to enable this feature are described in [an application packaging tutorial in Electron's documentation](http://electron.atom.io/docs/v0.36.0/tutorial/application-packaging/). When the value is `true`, pass default configuration to the `asar` module. The configuration values listed below can be customized when the value is an `Object`. Supported parameters include, but are not limited to:
+- `filename` (*String*): circumvents the asar generation and lets you specify your own pre-built asar file/directory.
```
malept: The description should start with "A path to", like `ordering` below it.

test/_setup.js
```diff
@@ -67,13 +67,14 @@ function npmInstallForFixture (fixture) {
         return true
       } else {
         console.log(`Running npm install in fixtures/${fixture}...`)
-        return exec('npm install --no-bin-links', {cwd: fixtureDir})
+        return exec('npm install --no-bin-links --no-package-lock', {cwd: fixtureDir})
```
malept: This should be in a different PR.

test/asar.js
```diff
@@ -53,3 +53,23 @@ util.testSinglePlatform('asar test', (t, opts) => {
...
+  return util.packageAndEnsureResourcesPath(t, opts)
+    .then(generatedResourcesPath => {
+      resourcesPath = generatedResourcesPath
+      return fs.stat(path.join(resourcesPath, 'app.asar'))
+    }).then(stats => {
+      t.true(stats.isFile(), 'app.asar should exist under the resources subdirectory when opts.asar.filename points to a prebuilt asar')
+      return fs.pathExists(path.join(resourcesPath, 'app'))
+    }).then(exists => {
+      return t.false(exists, 'app subdirectory should NOT exist when app.asar is built')
+    })
```
malept: This test should check that the copied asar is the same as the input asar. There is a utility function to assert this, I believe.

test/fixtures/asar-prebuilt/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "PrebuiltAsar",
+  "version": "0.0.1",
+  "description": "A test of the renamed from `electron-prebuilt` to `electron`",
```
malept: The description needs to be updated.

platform.js
```diff
@@ -154,6 +154,19 @@ class App {
 
     const dest = path.join(this.originalResourcesDir, 'app.asar')
     debug(`Running asar with the options ${JSON.stringify(asarOptions)}`)
+
+    if (asarOptions.filename) {
```
malept: I think that if there are other `asarOptions` other than `filename` and `filename` is set, it should warn that the other options won't be used.

platform.js
```diff
@@ -154,6 +154,19 @@ class App {
...
+      const src = path.resolve(asarOptions.filename)
+
+      return fs.stat(src)
+        .then(stat => {
+          const target = stat.isFile() ? dest : this.originalResourcesDir
+          debug(`Copying asar: ${src} to ${target}`)
+          return fs.copy(src, target, {overwrite: false, errorOnExist: true})
+        })
+        .then(() => fs.remove(this.originalResourcesAppDir))
```
malept: I think for efficiency purposes, if `asar.filename` is set, we shouldn't even copy the app directory in the first place. Then we can get rid of this `fs.remove` call.

platform.js
```diff
@@ -154,6 +154,19 @@ class App {
 
     const dest = path.join(this.originalResourcesDir, 'app.asar')
     debug(`Running asar with the options ${JSON.stringify(asarOptions)}`)
+
+    if (asarOptions.filename) {
+      const src = path.resolve(asarOptions.filename)
+
+      return fs.stat(src)
+        .then(stat => {
+          const target = stat.isFile() ? dest : this.originalResourcesDir
```
malept: Could you explain how `target` is determined? Shouldn't target always be `dest`?

test/fixtures/asar-prebuilt/package.json
```diff
@@ -0,0 +1,10 @@
...
+  "main": "main.js",
+  "devDependencies": {
+    "electron": "^1.8.0",
+    "electron-prebuilt": "^1.4.0"
```
malept: There should only be one item in `devDependencies`, and it should be `"electron": "1.3.1"`. Every time we add a new Electron version to the tests, it takes longer to run the tests due to downloading prebuilt binaries.

.editorconfig
jsg2021: sorry about that, I didn't mean to add that. I was trying to follow the current style and that file made my editor honor it.

platform.js
jsg2021: if the user specified a directory (containing their app.asar and additional resources) fs-extra's `copy` requires the dest to be a directory... if the src is a file, it requires dest to be a file.

platform.js
jsg2021: should that happen here? or in the argument parsing?

platform.js
jsg2021: I started to go down that route, but tried to keep the change minimal. I'll short circuit that.

platform.js
malept: It should happen here. If it happens in argument parsing, people who use the JS API (such as users of Electron Forge or ember-electron) will not get the warning.

platform.js
malept: It's not clear from the docs that you can specify a directory for `asar.filename`. Although even if it did, I would disagree with it because the name of the option makes that choice not obvious. I think the value of `asar.filename` should only be an ASAR file.

NEWS.md
```diff
@@ -4,6 +4,10 @@
 
 [Unreleased]: https://github.com/electron-userland/electron-packager/compare/v12.0.0...master
 
+### Added
+
+* `asar.filename` to support pre-built asar files.
```
malept: Minor nit, remove the period and add ` (#823)` at the end.

docs/api.md
```diff
@@ -183,6 +183,7 @@ is not restricted to the official set if [`download.mirror`](#download) is set.
 *Boolean* or *Object* (default: `false`)
 
 Whether to package the application's source code into an archive, using [Electron's archive format](https://github.com/electron/asar). Reasons why you may want to enable this feature are described in [an application packaging tutorial in Electron's documentation](http://electron.atom.io/docs/v0.36.0/tutorial/application-packaging/). When the value is `true`, pass default configuration to the `asar` module. The configuration values listed below can be customized when the value is an `Object`. Supported parameters include, but are not limited to:
+- `filename` (*String*): A path to a pre-built asar file. This will circumvent the asar generation and lets you specify your own pre-built asar file.
```
malept: This is kind of redundant. Let me think about a better way to word this.

platform.js
```diff
@@ -99,26 +99,35 @@ class App {
...
-          this.originalResourcesAppDir,
-          this.opts.electronVersion,
-          this.opts.platform,
-          this.opts.arch
-        ])
-      } else {
-        return true
-      }
-    })
+    const asarOptions = common.createAsarOpts(this.opts) || {}
```
malept: At this point `asarOptions` should be an instance variable, no need to calculate it twice.

platform.js
```diff
@@ -99,26 +99,35 @@ class App {
...
+
+    // don't copy the app if the asar filename is specified.
+    const stage = (asarOptions.filename)
+      ? Promise.resolve()
+      : fs.copy(this.opts.dir, this.originalResourcesAppDir, {
+        filter: ignore.userIgnoreFilter(this.opts),
+        dereference: this.opts.derefSymlinks
+      })
```
malept: I would rather have a real `if` statement than a multi-line ternary statement.

platform.js
```diff
@@ -99,26 +99,35 @@ class App {
...
+          return hooks.promisifyHooks(this.opts.afterPrune, [
+            this.originalResourcesAppDir,
+            this.opts.electronVersion,
+            this.opts.platform,
+            this.opts.arch
+          ])
+        } else {
+          return true
+        }
+      })
```
malept: I don't know whether afterPrune and afterCopy hooks should be run, since neither of those operations actually took place, and `this.originalResourcesAppDir` doesn't actually exist.

jsg2021: Yeah, I didn't what those hooks were for... I've changed it to just return early.

test/asar.js
```diff
@@ -53,3 +53,26 @@ util.testSinglePlatform('asar test', (t, opts) => {
...
+      return fs.stat(path.join(resourcesPath, 'app.asar'))
+    }).then(stats => {
+      t.true(stats.isFile(), 'app.asar should exist under the resources subdirectory when opts.asar.filename points to a prebuilt asar')
+      return util.areFilesEqual(opts.asar.filename, path.join(resourcesPath, 'app.asar'))
+  }).then(eql => {
+      t.true(eql, 'app.asar should equal the prebuilt asar')
+      return fs.pathExists(path.join(resourcesPath, 'app'))
+    }).then(exists => {
+      return t.false(exists, 'app subdirectory should NOT exist when app.asar is built')
+    })
```
malept: This can be:

```js
      }).then(exists => t.false(exists, 'app subdirectory should NOT exist when app.asar is built'))
```

platform.js
```diff
@@ -147,14 +153,31 @@ class App {
...
     const dest = path.join(this.originalResourcesDir, 'app.asar')
-    debug(`Running asar with the options ${JSON.stringify(asarOptions)}`)
-    return pify(asar.createPackageWithOptions)(this.originalResourcesAppDir, dest, asarOptions)
+    debug(`Running asar with the options ${JSON.stringify(this.asarOptions)}`)
+
+    if (this.asarOptions.filename) {
+      const src = path.resolve(this.asarOptions.filename)
+      if (Object.keys(this.asarOptions).length > 1) {
+        common.warning('asar.filename has been specified, all other asar options will be ignored')
+      }
```
malept: This should have a test.

platform.js
malept: I don't think returning early is the right solution... we may have to adjust the hook signature so that it's either the path to the resources app directory, or the path to the `app.asar` file (which would be a breaking change).

@MarshallOfSound since you have dealt with a lot of code relating to the hooks, you might have some insight here.

platform.js
jsg2021: what do you mean? Test that it warns? or test that the other options are ignored? If the later, how should that be done?

malept: Test that it warns. This is an example of doing that (although it should probably be genericized into a function):

https://github.com/electron-userland/electron-packager/blob/a7437b908e68e97b288173759d0538efe7798c82/test/mas.js#L18-L33

jsg2021: done.

malept:
I'm also a little worried about the drop in code coverage, but let's see what it looks like after the latest changes go through CI.

jsg2021:
I updated the usage.txt.

platform.js
```diff
@@ -147,14 +153,31 @@ class App {
...
+
+      return fs.stat(src)
+        .then(stat => {
+          if (!stat.isFile()) {
+            throw new Error(`${src} must be an asar file.`)
```
malept: This needs a test.

jsg2021: done

docs/api.md
```diff
@@ -183,6 +183,7 @@ is not restricted to the official set if [`download.mirror`](#download) is set.
 *Boolean* or *Object* (default: `false`)
 
 Whether to package the application's source code into an archive, using [Electron's archive format](https://github.com/electron/asar). Reasons why you may want to enable this feature are described in [an application packaging tutorial in Electron's documentation](http://electron.atom.io/docs/v0.36.0/tutorial/application-packaging/). When the value is `true`, pass default configuration to the `asar` module. The configuration values listed below can be customized when the value is an `Object`. Supported parameters include, but are not limited to:
+- `filename` (*String*): A path to a pre-built asar file. (This will circumvent the asar generation)
```
malept: > A path to a pre-built asar file (this will circumvent asar generation).

test/asar.js
```diff
@@ -53,3 +54,40 @@ util.testSinglePlatform('asar test', (t, opts) => {
...
+  opts.asar = {
+    'unpack': '*.pac',
+    'unpackDir': 'dir_to_unpack',
+    'filename': path.join(opts.dir, 'app.asar')
+  }
+  let resourcesPath
+  sinon.spy(console, 'warn')
+  return util.packageAndEnsureResourcesPath(t, opts)
+    .then(generatedResourcesPath => {
+      console.warn.calledWithExactly('WARNING: asar.filename has been specified, all other asar options will be ignored')
```
malept: Shouldn't this be wrapped in `t.true()`?

usage.txt
```diff
@@ -25,6 +25,7 @@ asar               whether to package the source code within your app into an ar
                    --asar and its sub-properties simultaneously.
 
                    Properties supported:
+                   - filename: path to a pre-built asar file. (This will circumvent the asar generation)
```
malept: > path to a pre-built asar file (this will circumvent asar generation)

test/asar.js
jsg2021: sorry, i’ve been in jest land for a while, i assumed sinon made the assertion.  i’ll update.

usage.txt
jsg2021: done

test/asar.js
jsg2021: done

docs/api.md
jsg2021: done

jsg2021:
None of this is breaking existing configurations. Without adding the `asar.filename` option, nothing changes.

The whole point of this option is to take control of the app bundle generation... the hooks can be external to this once the filename is used... if something is needed to be done to the app package, a more general hook can be added?

jsg2021:
the after prune hook still needs to be called since its unrelated to copy.

MarshallOfSound:
RE the hooks issue I believe the operation should throw an error if `afterCopy` or `afterPune` hooks are provided and a prebuilt ASAR file is provided , these are incompatible options.

Also I don't think this option should go into the `options.asar` option.  That object is typically sent pretty much through to the ASAR module.  This goes nowhere near the ASAR module, I'd prefer this if it was something like `options.prebuiltAsar`.  This should also throw errors if that option is used in conjunction with any other ASAR option.

Also @malept we should just make forge error out instantly if this option is used 😄

malept:
The docs should be updated to note the `asar`, `afterCopy`, and `afterPrune` incompatibilities.

jsg2021:
I think I got all the suggestions implemented.

common.js
```diff
@@ -43,6 +43,25 @@ function parseCLIArgs (argv) {
     args.out = null
   }
 
+  if (args.prebuiltAsar) {
+    if (args.asar) {
+      warning('--prebuiltAsar and --asar options are incompatible. Ignoring --asar')
+      delete args.asar
+    }
+    console.log('\n\n%s\n\n', JSON.stringify(args, null, 2))
```
malept: ✂️  debug line

common.js
```diff
@@ -43,6 +43,25 @@ function parseCLIArgs (argv) {
...
+    if (!args.prune) {
+      warning('--prebuiltAsar and --no-prune are incompatible. Ignoring --no-prune')
+    }
+
+    if (args.ignore) {
+      warning('--prebuiltAsar and --ignore are incompatible. Ignoring --ignore')
+    }
+
+    if (!args['deref-symlinks']) {
```
malept: Should be `!args.derefSymlinks`

common.js
```diff
@@ -43,6 +43,25 @@ function parseCLIArgs (argv) {
...
+      warning('--prebuiltAsar and --no-deref-symlinks are incompatible. Ignoring --no-deref-symlinks')
+    }
+  }
+
```
malept: This should be outside of the CLI argument parsing, as I mentioned yesterday, because it will not give warnings to consumers of the JavaScript API.

docs/api.md
```diff
@@ -67,6 +67,10 @@ packager({
 })
 ```
 
+**NOTE:** `afterCopy` will not be called if `prebuiltAsar` is set.
+
+---
```
malept: The lines are unnecessary.

docs/api.md
```diff
@@ -327,6 +336,14 @@ Arbitrary combinations of individual platforms are also supported via a comma-de
 The non-`all` values correspond to the platform names used by [Electron releases]. This value
 is not restricted to the official set if [`download.mirror`](#download) is set.
 
+##### `prebuiltAsar`
+
+*String*: (default: `undefined`)
+
+A path to a pre-built asar file. 
+
+**NOTE:** This option circumvents many options such as `asar`, `afterCopy`, `afterPrune`, `ignore`, `prune`, or any option related to the copy processes. These options are for generating an app directory/asar with this tool. However, if a prebuilt asar file is provided, these options are not needed.
```
malept: I'd get rid of `, or any option related to the copy processes`. Also, it's a little too verbose in general, but I need some time to think of a better way to word it.

platform.js
```diff
@@ -101,30 +101,34 @@ class App {
 
   copyTemplate () {
     // don't copy the app if the asar filename is specified.
-    if ((this.asarOptions || {}).filename) {
+    if (this.opts.prebuiltAsar) {
+      if (this.opts.afterCopy) {
+        return Promise.reject(new Error('afterCopy is incompatible with prebuiltAsar'))
```
malept: You should just be able to `throw new Error`.

docs/api.md
jsg2021: I thought it was a nice visual grouping. I'll drop them.

malept: I think if we're going to do that, it should be in a separate PR.

platform.js
jsg2021: This isn't an async function, I was trying to make as little of a splash as possible. I hadn't looked to see if the callers setup to handle thrown exceptions and not just promise rejections.

jsg2021:
I've updated to use your suggested initialize()

jsg2021:
looks like a transient error caused the signing test to fail for one of the jobs... but the rest seem to be passing. I haven't looked further into it, since its outside the scope of my changes and it has been passing.

jsg2021:
Anything I should change?

jsg2021:
@malept Would you like anything else? or do you think this is good?

NEWS.md
```diff
@@ -24,6 +24,10 @@
 
 * Upgraded `galactus` to `^0.2.1` to fix a bug with relative paths
 
+### Added
```
MarshallOfSound: This is now in the wrong place

jsg2021: ah, should be fixed now

jsg2021:
I hope all is well? Is this still wanted?

jsg2021:
Your cleanup looks way better than my changes. :) I didn't want to change too much without fully comprehending the codebase. Sorry if I misunderstood your requests. Hope this gets merged 👍

jsg2021:
I've been using this branch to bundle my apps, the flag works well. Hope we can merge soon.

jsg2021:
Is this blocked by something? Anything I could help with?

jsg2021:
🎉

Issue #831, opened 2018-04-25
Cannot run with Node 10


mattridley:
When packaging my electron app with Node 10 I get the following error in the console.

CANNOT RUN WITH NODE 10.0.0
Electron Packager requires Node 4.0 or above.

In cli.js the check evaluates to [10, 0, 0] < [4, 0, 0] which is true, whereas previously [9, 0, 0] returns as false.

I'll raise a PR now to simplify this check so it continues to work with newer versions of node.

Joni4Games:
Any solution yet? Do we have to downgrade to node 4.x?

Triforcey:
@Joni4Games It is patched. Just use the latest 9.x release until the fix is published.

malept:
Thanks for your patience, a fix has been released in 12.0.2.

psharneja:
well, i'm using version 12.1.0 and issue remains the same

malept:
@psharneja I cannot reproduce this. My steps (on Linux):

```
# step 0: install nodeenv <https://github.com/ekalinin/nodeenv>
nodeenv -n 10.2.1 node10
source node10/bin/activate
git clone https://github.com/electron-userland/electron-packager
cd electron-packager
npm install
./cli.js --version
```
My output:
```
Electron Packager 12.1.0
Node v10.2.1
Host Operating system: linux (x64)
```

Triforcey:
On my system the issue also appears to be fixed in 12.1.0.

jsejcksn:
I am beginning to learn about Electron through the [electron-api-demos](https://github.com/electron/electron-api-demos) app. When trying to run `npm run-script package` I also get this error.

I am running the following versions:
 - Node: 10.4.1
 - electron-packager: 12.1.0
 - macOS 10.13.5.

malept:
@jsejcksn if you run `npm run package -- --version` and it doesn't immediately print out the Electron Packager version and exit, you don't have Electron Packager 12.1.0 installed. In that case, run `npm update`.

jsejcksn:
Hi @malept: Thanks for responding. Running your example command produces an error:

```
$ npm run-script package -- --version

> ElectronAPIDemos@1.3.0 package /path/to/electron-api-demos
> npm-run-all package:* "--version"

ERROR: Invalid Option: --version
```

[...]

PR #832, opened 2018-04-25
#831 Update node version check to support node 10

mattridley:
Update the node version check to support node 10 (#831). The current check fails once the node version hits double figures.

This PR simplifies the check to only compare the major version number.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -2,8 +2,8 @@
 
 'use strict'
 
-var nodeVersionInfo = process.versions.node.split('.').map(function (n) { return Number(n) })
-if (nodeVersionInfo < [4, 0, 0]) {
+var nodeMajorVersion = Number(process.versions.node.split('.')[0])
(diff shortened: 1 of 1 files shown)


cli.js
```diff
@@ -2,8 +2,8 @@
 
 'use strict'
 
-var nodeVersionInfo = process.versions.node.split('.').map(function (n) { return Number(n) })
-if (nodeVersionInfo < [4, 0, 0]) {
+var nodeMajorVersion = Number(process.versions.node.split('.')[0])
+if (nodeMajorVersion < 4) {
```
jsg2021: If you want to check if the version is greater than some specific semver in the future, I would require `semver` and use its comparison method... inequality checks on arrays convert to strings ... so `[4,2,0] > [4,10,0] == true` but should be `false` when comparing two versions.  

But, if all this cares about is the major version, then this approach is sufficient.

PR #953, opened 2019-02-20
Add .DS_Store to the default ignores list

brandonhorst:

**Summarize your changes:**
.DS_Store seems like a reasonable default to ignore.

### DIFF
--- a/ignore.js
+++ b/ignore.js
@@ -9,7 +9,8 @@ const targets = require('./targets')
 const DEFAULT_IGNORES = [
   '/\\.git($|/)',
   '/node_modules/\\.bin($|/)',
-  '\\.o(bj)?$'
+  '\\.o(bj)?$',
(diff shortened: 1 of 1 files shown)


malept:
I appreciate the idea, but this is getting to be a slippery slope. Why not add `.*.sw?` for Vim, `.nyc_output` for `nyc`, and a whole other host of files? I would much rather have something like [`ignore-by-default`](https://www.npmjs.com/package/ignore-by-default) that I don't have to manage. (But not that, because ignoring `node_modules` would be bad.)

malept:
While I'm thinking about it, I'd probably want to avoid adding `.gitignore`-style ignore file support, because users would have a tendency to do something like `--ignore.file=.gitignore`, which will end very badly for them, because it would then ignore `node_modules`.

brandonhorst:
Seems reasonable to me!

PR #957, opened 2019-03-05
fix(targets): Linux/ia32 is no longer supported for Electron >= 4

malept:

**Summarize your changes:**

See:

* https://electronjs.org/blog/linux-32bit-support
* https://github.com/electron/electron/issues/16407

### DIFF
--- a/targets.js
+++ b/targets.js
@@ -15,6 +15,7 @@ const officialPlatformArchCombos = {
 
 const linuxArchBuildVersions = {
   arm64: '>= 1.8.0',
+  ia32: '< 4.0.0-beta.1',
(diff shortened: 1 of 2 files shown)


Nikeron:
Linux only? Win32 will be alive? Are you seriously? 👏

malept:
All platform/arch decisions are up to the Electron group in charge of releases, Electron Packager only follows their direction.

In this case, I'm probably going to revert this PR and release a new version. My understanding is that a warning will be emitted from `electron-download` when 32-bit Linux builds >= 4.0.0 are downloaded, so it's at the developer's risk to use them. (They may or may not work.)

PR #984, opened 2019-04-25
Replace extract-zip with cross-zip

malept:

**Summarize your changes:**

`extract-zip` is effectively unmaintained, and we're already using `cross-zip` in Forge (for different reasons).

`cross-zip` also has fewer Node dependencies, although the tradeoff is that [on Windows it requires a certain version of .NET Framework & Powershell](https://github.com/feross/cross-zip/blob/v2.1.5/README.md#windows-users).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,7 +29,6 @@ feature. Known modules include:
 * `electron-download`
 * `electron-osx-sign`
 * `electron-packager` (always use this one before filing an issue)
-* `extract-zip`
 * `get-package-info`
 
 We use the [`debug`](https://www.npmjs.com/package/debug#usage) module for this functionality. It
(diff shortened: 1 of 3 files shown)


malept:
Well, we'll see how many complaints we get from this change :man_shrugging:

campersau:
Just want to note that `cross-zip` does not support symlinks on windows but because of this https://github.com/electron/electron-packager/issues/933 (https://github.com/jprichardson/node-fs-extra/issues/684) it does not really matter. Windows currently can't build macOS packages because of these symlinks issues...

malept:
@campersau this is confusing because both of the issues you reference were created *before* this PR was merged. If you can prove that you don't have this issue if you only replace `cross-zip` with `extract-zip` (and can also create a unit test to test for this), I will figure out a different way of extracting ZIP files.

malept:
For the record, this is still a problem:

> `extract-zip` is effectively unmaintained

PR #1033, opened 2019-07-24
fix: support Electron v6 helpers

MarshallOfSound:
Adds support for the new `Electron Helper (Renderer).app` and `Electron Helper (Plugin).app` helper bundles 👍

### DIFF
--- a/test/darwin.js
+++ b/test/darwin.js
@@ -169,6 +169,27 @@ async function appHelpersBundleTest (t, opts, helperBundleId, appBundleId) {
   return assertCFBundleIdentifierValue(t, helperNPObj, helperBundleIdentifier + '.NP', 'CFBundleName should reflect opts.helperBundleId, opts.appBundleId or fallback to default in helper NP app')
 }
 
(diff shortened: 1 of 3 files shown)


test/darwin.js
```diff
@@ -169,6 +169,33 @@ async function appHelpersBundleTest (t, opts, helperBundleId, appBundleId) {
...
 }
 
+async function appHelpersBundleElectron6Test (t, opts, helperBundleId, appBundleId) {
+  if (helperBundleId) {
+    opts.helperBundleId = helperBundleId
+  }
+  if (appBundleId) {
+    opts.appBundleId = appBundleId
+  }
+  opts.electronVersion = '6.0.0-beta.15'
```
malept: This Electron version needs to be added to the testcase setup. Although it needs to only download the prebuild `darwin` binary... :cry:

malept: I should be clearer here... We pre-download the Electron binaries used so that we don't get timeouts during the actual testsuite run.

MarshallOfSound: Added to the initial download list 👍

PR #1047, opened 2019-08-31
fix: send the properly formatted options from the CLI to electron-osx-sign

jhen0409:

**Summarize your changes:**

Fix osx-sign can't get correctly sub properties, just use `osx-sign` arg instead of `osxSign`.

For `yargs-parser`, use camel case of args is meaning the subprops will be also converted to camel case, but it will break the [`osx-sign`](https://github.com/electron/electron-osx-sign/blob/master/sign.js#L307) usage like `entitlements-inherit`.

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -67,6 +67,9 @@ module.exports = {
     if (args.osxSign === 'true') {
       warning('--osx-sign does not take any arguments, it only has sub-properties (see --help)')
       args.osxSign = true
+    } else if (typeof args.osxSign === 'object' && !Array.isArray(args.osxSign)) {
+      // Keep kebab case of sub properties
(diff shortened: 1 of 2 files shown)


src/mac.js
```diff
@@ -288,7 +288,7 @@ class MacApp extends App {
   }
 
   async signAppIfSpecified () {
-    const osxSignOpt = this.opts.osxSign
+    const osxSignOpt = this.opts['osx-sign'] || this.opts.osxSign
```
malept: This is in the wrong place, it should probably be in `src/cli.js`.

jhen0409: @malept OK, I made the change and added the test.

PR #1049, opened 2019-09-09
feat: add usageDescription option for darwin/mas targets

malept:

**Summarize your changes:**

Adds a convenience option to set usage descriptions for the macOS store, on `darwin`/`mas` targets.

Reference: https://github.com/electron/electron/pull/19871#pullrequestreview-278011263

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -474,7 +474,7 @@ Example:
 ```javascript
 {
   usageDescription: {
-    Camera: "Needed for video calls",
+    Camera: 'Needed for video calls',
     Microphone: 'Needed for voice calls'
   }
 }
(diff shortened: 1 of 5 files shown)


docs/api.md
```diff
@@ -456,6 +456,17 @@ Each *Object* is required to have the following properties:
   example, specifying `myapp` would cause URLs such as `myapp://path` to be opened with the app.
   Maps to the `CFBundleURLSchemes` metadata property.
 
+##### `usageDescription`
```
MarshallOfSound: It's not clear reading these docs that this should be an object

malept: Yep, I forgot the type. I added a type and an example in https://github.com/electron/electron-packager/pull/1049/commits/525c04d629e630b6f449e842b1569b18dd205268.

I really need to figure out a better way to deal with the API docs. Perhaps TypeScript definitions + typedoc.

PR #1074, opened 2019-11-14
Upgrade rcedit to support 64-bit

hoffmabc:
* [ X ] I have read the [contribution documentation](https://github.com/electron/electron-packager/blob/master/CONTRIBUTING.md) for this project.
* [ X ] I agree to follow the [code of conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project follows, as appropriate.

This simply upgrades rcedit package to 2.1.0 or above, which now supports 64-bit rcedit.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "junk": "^3.1.0",
     "parse-author": "^2.0.0",
     "plist": "^3.0.0",
-    "rcedit": "^2.0.0",
+    "rcedit": "^2.1.0",
     "resolve": "^1.1.6",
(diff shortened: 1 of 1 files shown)


malept:
Thanks, but this should not be necessary. `^2.0.0` implies that anything in the 2.x.y range (`>= 2.0.0, < 3.0.0`) is allowed as the version of `rcedit` to install. You can update your version by either running `yarn upgrade electron-packager` or `npm upgrade rcedit --depth=100`, depending on which package manager you're using.

PR #1084, opened 2019-11-24
Revert "fix(win32): handle missing wine64 binaries in addition to wine"

Welder23:
Reverts electron/electron-packager#1083


malept:
Can you explain why?

malept:
I'm going to close this for now, but if you can respond with your reasons for why you opened this pull request, I will take a look and consider reopening with the new context.

PR #1090, opened 2019-12-13
fix: make relative paths work with `dir` option

bendemboski:
When handling the `icon`, `extendInfo`, and `extraResource` options were interpreting relative paths as relative to the current working directory, rather than relative to the `dir` option.

### DIFF
--- a/src/mac.js
+++ b/src/mac.js
@@ -138,7 +138,7 @@ class MacApp extends App {
     }
 
     if (typeof propsOrFilename === 'string') {
-      const plist = await this.loadPlist(propsOrFilename)
+      const plist = await this.loadPlist(this.inputPath(propsOrFilename))
       return Object.assign(this.appPlist, plist)
     } else {
       return Object.assign(this.appPlist, propsOrFilename)
--- a/src/platform.js
+++ b/src/platform.js
@@ -81,6 +81,10 @@ class App {
     return this.relativeRename(this.electronBinaryDir, this.originalElectronName, this.newElectronName)
   }
 
+  inputPath (filePath) {
+    return path.resolve(this.opts.dir, filePath)
+  }
+
   /**
    * Performs the following initial operations for an app:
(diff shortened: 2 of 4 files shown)


bendemboski:
Hmm...you're right, relative paths being relative to the cwd does seem to make more sense from a CLI perspective, and also when using the `electron-packager` API directly.

Backing up to the problem I'm aiming to solve, I'm using `electron-forge`, so all my options are in a config file or `package.json` inside an `electron-forge` project. In this context, it's very surprising to invoke `electron-forge` from some other folder using the `dir` option to point it at my project, and have the paths in my config files be relative to wherever I invoked the command from, rather than relative to my project.

The workaround isn't hard -- put my config in a config file (rather than `package.json`) where I can use `__dirname` to dynamically generate absolute paths. But when using `electron-forge`, this is surprising and kinda difficult to discover.

I guess potential options are:

1. Make changes along the lines of this PR, making the CLI interface a little counter-intuitive (agree, not a great option)
2. Leave as-is but update the documentation to call out this oddity when using the `dir` option, and maybe more carefully catch `ENOENT`s for files like this and have the error message point to the documentation (not bad, but intuitive interfaces are always better than relying on documentation)
3. Branch the logic somehow so relative paths coming from the CLI and relative paths coming from an `electron-forge` config file are treated differently.

It seems like (3) could be workable, but would require a little trickery. I don't think we really want `electron-forge` having knowledge of every path that might be passed to `electron-packager` and converting it to absolute based on the `dir` option over there. Maybe `electron-packager` could expose an API option that instructs it to treat relative paths as relative to `dir` rather than `cwd()`, and then `electron-forge` could set that by default (overridable in the config file for super-advanced users)?

Thoughts?

ChrisAntaki:
Option 2 seems like a good place to start. Perhaps the packager could warn users if they are passing a relative path for a BrowserWindow icon too (#1150)

PR #1099, opened 2020-01-11
fix: Ignore node_modules/.bin in all directories

vladh:

**Summarize your changes:**

`/node_modules/.bin` is excluded from the build by default, but I also had the following directories in my project:

* `node_modules/mocha/node_modules/.bin`
* `my-submodule/node_modules/.bin`
* `my-submodule/node_modules/mocha/node_modules/.bin`

This led to the following issue on packaging: https://github.com/electron-userland/electron-builder/issues/4559

I've modified the pattern to exclude `node_modules/.bin` directories across the project.

Note: this change actually matches the existing documentation in `api.md`.

### DIFF
--- a/src/ignore.js
+++ b/src/ignore.js
@@ -9,7 +9,7 @@ const targets = require('./targets')
 
 const DEFAULT_IGNORES = [
   '/\\.git($|/)',
-  '/node_modules/\\.bin($|/)',
+  'node_modules/\\.bin($|/)',
   '\\.o(bj)?$'
 ]
 


malept:
So I took a crack at writing a regression test in `test/ignore.js` and I can't get it to fail in master:

```javascript
test('ignore default: node_modules/.bin', util.testSinglePlatform(async (t, opts) => {
  opts.dir = util.fixtureSubdir('basic')
  opts.junk = false
  opts.prune = false

  const mkdirpBin = path.join('node_modules', 'extract-zip', 'node_modules', '.bin', 'mkdirp')
  t.true(await fs.pathExists(path.join(opts.dir, mkdirpBin)))
  const targetDir = await copyDirToTempDirWithIgnores(t, opts)
  await assertFileIgnored(t, targetDir, mkdirpBin)
}))
```

In order to debug the problem further, we need a minimal testcase to reproduce your problem. Using the [`electron-quick-start`](https://github.com/electron/electron-quick-start) repository as a base, could you please create a minimal Electron app that illustrates the issue you described, and post a link to it here?

malept:
Since the original issue reporter has not responded to the request for additional information in about a week, I'm going to assume that they have decided not to pursue this issue. If I'm incorrect, they can respond with answers to the questions posed, and I'll reopen.

PR #1149, opened 2020-05-18
Make 11.x work with Electron 6 and above

duyluongtruong:
This PR picks up changes from master which supports Electron 6x: 

To make version 11.x works with Electron 6.x (and above). We need that because we are still using electron-forge version 5.2.4 and electron-forge internally using electron-packager 11.x

### DIFF
--- a/mac.js
+++ b/mac.js
@@ -101,11 +101,15 @@ class MacApp extends App {
     })
   }
 
-  updateHelperPlist (base, suffix) {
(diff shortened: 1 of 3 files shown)


malept:
I'm not inclined to merge this as I don't have the bandwidth to maintain multiple versions of Packager. As an alternative, you can use [yarn resolutions](https://classic.yarnpkg.com/en/docs/selective-version-resolutions/) to force Electron Forge v5 use Packager v14.2.1 (or whatever version you want).

PR #1182, opened 2020-10-03
fix(copy-filter): add package manager lockfiles to default ignore

emab:
* [X] I have read the [contribution documentation](https://github.com/electron/electron-packager/blob/master/CONTRIBUTING.md) for this project.
* [X] I agree to follow the [code of conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project follows, as appropriate.
* [X] The changes are appropriately documented (if applicable).
* [X] The changes have sufficient test coverage (if applicable).
* [X] The testsuite passes successfully on my local machine (if applicable).

**Summarize your changes:**

Added `package-lock.json` and `yarn.lock` to the default ignored file regex.

### DIFF
--- a/src/copy-filter.js
+++ b/src/copy-filter.js
@@ -8,8 +8,8 @@ const prune = require('./prune')
 const targets = require('./targets')
 
 const DEFAULT_IGNORES = [
-  '/package-lock\\.json/',
-  '/yarn\\.lock/',
+  'package-lock\\.json$',
+  'yarn\\.lock$',
   '/\\.git($|/)',
   '/node_modules/\\.bin($|/)',
   '\\.o(bj)?$'
--- a/src/copy-filter.js
+++ b/src/copy-filter.js
@@ -8,8 +8,8 @@ const prune = require('./prune')
 const targets = require('./targets')
 
 const DEFAULT_IGNORES = [
-  'package-lock\\.json$',
-  'yarn\\.lock$',
+  '\bpackage-lock\\.json$',
+  '\byarn\\.lock$',
   '/\\.git($|/)',
   '/node_modules/\\.bin($|/)',
   '\\.o(bj)?$'
--- a/src/copy-filter.js
+++ b/src/copy-filter.js
@@ -8,6 +8,8 @@ const prune = require('./prune')
 const targets = require('./targets')
 
 const DEFAULT_IGNORES = [
+  '/package-lock\\.json/',
+  '/yarn\\.lock/',
   '/\\.git($|/)',
   '/node_modules/\\.bin($|/)',
   '\\.o(bj)?$'
--- a/src/copy-filter.js
+++ b/src/copy-filter.js
@@ -8,8 +8,8 @@ const prune = require('./prune')
 const targets = require('./targets')
 
 const DEFAULT_IGNORES = [
-  '\bpackage-lock\\.json$',
-  '\byarn\\.lock$',
+  '(/|\b)package-lock\\.json$',
+  '(/|\b)yarn\\.lock$',
   '/\\.git($|/)',
   '/node_modules/\\.bin($|/)',
   '\\.o(bj)?$'
--- a/src/copy-filter.js
+++ b/src/copy-filter.js
@@ -8,8 +8,8 @@ const prune = require('./prune')
 const targets = require('./targets')
 
 const DEFAULT_IGNORES = [
-  '(/|\b)package-lock\\.json$',
-  '(/|\b)yarn\\.lock$',
+  '/package-lock\\.json$',
+  '/yarn\\.lock$',
   '/\\.git($|/)',
   '/node_modules/\\.bin($|/)',
   '\\.o(bj)?$'
--- a/test/copy-filter.js
+++ b/test/copy-filter.js
@@ -96,6 +96,12 @@ test('ignore: not junk when junk: false', util.testSinglePlatform(async (t, opts
   const targetDir = await copyDirToTempDirWithIgnores(t, opts)
   await assertFileNotIgnored(t, targetDir, 'subfolder/Thumbs.db')
 }))
+test('ignore: package manager lock files', util.testSinglePlatform(async (t, opts) => {
+  opts.dir = t.context.workDir
+  const targetDir = await copyDirToTempDirWithIgnores(t, opts)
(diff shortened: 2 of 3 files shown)


src/copy-filter.js
```diff
@@ -8,6 +8,8 @@ const prune = require('./prune')
 const targets = require('./targets')
 
 const DEFAULT_IGNORES = [
+  '/package-lock\\.json/',
+  '/yarn\\.lock/',
```
malept, line 12: In the very unlikely event that this is part of a larger filename...
```suggestion
  '/package-lock\\.json$/',
  '/yarn\\.lock$/',
```

test/copy-filter.js
```diff
@@ -96,6 +96,12 @@ test('ignore: not junk when junk: false', util.testSinglePlatform(async (t, opts
   const targetDir = await copyDirToTempDirWithIgnores(t, opts)
   await assertFileNotIgnored(t, targetDir, 'subfolder/Thumbs.db')
 }))
+test('ignore: package manager lock files', util.testSinglePlatform(async (t, opts) => {
```
malept, line 99: For consistency with the other default ignore tests:

```suggestion
test('ignore default: package manager lock files', util.testSinglePlatform(async (t, opts) => {
```

test/copy-filter.js
```diff
@@ -96,6 +96,12 @@ test('ignore: not junk when junk: false', util.testSinglePlatform(async (t, opts
...
+  opts.dir = t.context.workDir
+  const targetDir = await copyDirToTempDirWithIgnores(t, opts)
+  await assertFileIgnored(t, targetDir, 'package-lock.json')
+  await assertFileIgnored(t, targetDir, 'yarn.lock')
+}))
```
malept, line 104: Can you move this adjacent to the `.o`/`.obj` tests so it's easier to find?

test/copy-filter.js
```diff
@@ -96,6 +96,12 @@ test('ignore: not junk when junk: false', util.testSinglePlatform(async (t, opts
   const targetDir = await copyDirToTempDirWithIgnores(t, opts)
   await assertFileNotIgnored(t, targetDir, 'subfolder/Thumbs.db')
 }))
+test('ignore: package manager lock files', util.testSinglePlatform(async (t, opts) => {
+  opts.dir = t.context.workDir
+  const targetDir = await copyDirToTempDirWithIgnores(t, opts)
+  await assertFileIgnored(t, targetDir, 'package-lock.json')
+  await assertFileIgnored(t, targetDir, 'yarn.lock')
```
malept, line 103: I don't think this test does what you want? As far as I know, `t.context.workDir` doesn't contain any package lock files before the copy happens. I suggest adding an empty `package-lock.json` and `yarn.lock` to a new fixture folder (`ignore-package-lock` perhaps, making sure they're added to git with `git add --force`) and then setting `opts.dir` to that path.

malept (changes requested):
Thanks for working on this PR! This approach generally works, but I have some suggestions that will make it more robust that I'd like to get in before we merge this (mainly regarding the test that was added).

src/copy-filter.js
emab: I initially went for `^/package-lock\\.json$`, would that help here or is end of match more important than start>

test/copy-filter.js
emab: Thanks for helping me out on my first open source issue! Yeah looking back this isn't testing what I thought it was. Changing the filename allows the test to pass regardless of filename.

I can see the code to add/set a fixture folder, but I can't see how you'd add files to that folder via `git add` in order to test that they aren't included.

malept: Sorry if that was unclear, I meant add empty (fake) package manager lock files to a new fixture folder via `git add --force` and then commit it to the branch, instead of trying to do that sort of thing via test code.

emab: Nope looking back that makes perfect sense - I misunderstood. I'll get those changes put in asap.

src/copy-filter.js
emab: Added

test/copy-filter.js
emab: Renamed

test/copy-filter.js
emab: Moved this next to other `ignore: default` tests

test/copy-filter.js
emab: Added a fixture folder with lockfiles, added to git and fixed up the test.

src/copy-filter.js
```diff
@@ -8,6 +8,8 @@ const prune = require('./prune')
 const targets = require('./targets')
 
 const DEFAULT_IGNORES = [
+  'package-lock\\.json$',
+  'yarn\\.lock$',
```
malept, line 12: One small suggestion:

```suggestion
  '\bpackage-lock\\.json$',
  '\byarn\\.lock$',
```

src/copy-filter.js
```diff
@@ -8,8 +8,8 @@ const prune = require('./prune')
 const targets = require('./targets')
 
 const DEFAULT_IGNORES = [
-  'package-lock\\.json$',
-  'yarn\\.lock$',
+  '\bpackage-lock\\.json$',
```
emab, line 11:
```suggestion
  '(/|\b)package-lock\\.json$',
  '(/|\b)yarn\\.lock$',
```

This would probably cover all cases?

emab:
Sorry about the force push - I pushed a commit from the wrong git account.

src/copy-filter.js
malept: I think adding the slash is redundant. `\b` looks for boundaries, of which `/` creates one.

src/copy-filter.js
```diff
@@ -8,6 +8,8 @@ const prune = require('./prune')
 const targets = require('./targets')
 
 const DEFAULT_IGNORES = [
+  '(/|\b)package-lock\\.json$',
+  '(/|\b)yarn\\.lock$',
```
malept, line 12: Hmm. I just saw the test failure. I think this is a better pattern:
```suggestion
  '/package-lock\\.json$',
  '/yarn\\.lock$',
```

emab: Yeah - this should cover any lock files in any directory.

malept (approved):
Looks good!

PR #1191, opened 2020-11-15
fix(targets): allow checking official builds against prerelease versions

### LINKED ISSUE #1190: Building for Linux arm64 with a beta version of Electron doesn't work
### Preflight Checklist

### Issue Details

* **Electron Packager Version:** 15.1.0
* **Electron Version:** 11.0.0-beta.22
* **Operating System:** Ubuntu 20.04
* **Last Known Working Electron Packager version:**:

I'm trying to create a Linux Arm64 package for Electron 11.0.0-beta.22, but when I call electron-packager from my build script, it returns immediately with an empty paths array, and I get the following error:
```
WARNING: official linux/arm64 support only exists in Electron >=1.8.0
```
Even though I'm using an Electron version greater than that.

Looking through the code, I think this is ultimately an issue in [officialBuildExists](https://github.com/electron/electron-packager/blob/master/src/targets.js#L81-L83), which is returning false in this case when it shouldn't. If I try calling semver:
```
semver.satisfies('11.0.0-beta.22', '>= 1.8.0')
```
I get `false`, which seems wrong. (If I test it with "11.0.0", it works correctly). That causes the packager script to assume that no builds are available and return early without actually packaging anything.

I'm not sure what the right fix is here - is this an issue with the `semver` package? Or is there a different check that's supposed to be used here? But if you have an idea what the right fix should be, I'd be happy to make a PR for it.

PalmerAL:

**Summarize your changes:**

Fixes #1190 

The behavior exhibited by semver in that issue is apparently [intended behavior](https://github.com/npm/node-semver#prerelease-tags), but I don't think it makes much sense in this context, so I've set the option to disable it, which allows ARM64 builds to work as intended.

### DIFF
--- a/src/targets.js
+++ b/src/targets.js
@@ -79,7 +79,7 @@ function validOfficialPlatformArch (opts, platform, arch) {
 }
 
 function officialBuildExists (opts, buildVersion) {
-  return semver.satisfies(opts.electronVersion, buildVersion)
+  return semver.satisfies(opts.electronVersion, buildVersion, { includePrerelease: true })
 }
 
 function allPlatformsOrArchsSpecified (opts) {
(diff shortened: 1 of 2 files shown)


malept (approved):
Thanks for the fix!

PR #1233, opened 2021-04-01
feat: allow extension of helper app plists with extendHelperInfo

MarshallOfSound:
For apps that want to customize the Helper app plist files they currently have to do so in an `afterCopy` or other out-of-band step.  Adding `extendHelperInfo` to match `extendInfo` makes it much easier to extend these plist files.

This PR also ensures we set `CFBundleShortVersionString` and `CFBundleVersion` on helper plists.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,6 +14,10 @@
 * Support for customizing Windows targets on darwin/arm64 (#1260)
 * Support for customizing Windows targets on WSL without Wine installed (#1260)
 
+### Added
+
+* Add `extendHelperInfo` option to allow extend helper app `Info.plist` files
+
 ## [15.2.0] - 2020-12-04
 
 [15.2.0]: https://github.com/electron/electron-packager/compare/v15.1.0...v15.2.0
--- a/NEWS.md
+++ b/NEWS.md
@@ -16,7 +16,7 @@
 
 ### Added
 
-* Add `extendHelperInfo` option to allow extend helper app `Info.plist` files
+* `extendHelperInfo` option to allow extending helper app `Info.plist` files
 
 ## [15.2.0] - 2020-12-04
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,10 @@
 
 [Unreleased]: https://github.com/electron/electron-packager/compare/v15.3.0...main
 
+### Added
+
+* `extendHelperInfo` option to allow extending helper app `Info.plist` files (#1233)
+
 ## [15.3.0] - 2021-07-17
 
 [15.3.0]: https://github.com/electron/electron-packager/compare/v15.2.0...v15.3.0
@@ -13,10 +17,6 @@
 * Bundled app validation to ensure that both `package.json` and the main entry point exist (#1257)
 * Support for customizing Windows targets on darwin/arm64 (#1260)
 * Support for customizing Windows targets on WSL without Wine installed (#1260)
-
-### Added
-
-* `extendHelperInfo` option to allow extending helper app `Info.plist` files
 
 ## [15.2.0] - 2020-12-04
 
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -334,6 +334,19 @@ declare namespace electronPackager {
      */
     extendInfo?: string | { [property: string]: any }; // eslint-disable-line @typescript-eslint/no-explicit-any
     /**
+     * When the value is a string, specifies the filename of a `plist` file. Its contents are merged
+     * into all the Helper app's `Info.plist` files.
+     * When the value is an `Object`, it specifies an already-parsed `plist` data structure that is
+     * merged into all the Helper app's `Info.plist` files.
+     *
+     * Entries from `extendHelperInfo` override entries in the helpers apps `Info.plist` file supplied by
+     * `electron`, `electron-prebuilt-compile`, or `electron-prebuilt`, but are overridden by other
+     * options such as [[appVersion]] or [[appBundleId]].
+     *
+     * @category macOS
+     */
+    extendHelperInfo?: string | { [property: string]: any }; // eslint-disable-line @typescript-eslint/no-explicit-any
+    /**
      * One or more files to be copied directly into the app's `Contents/Resources` directory for
      * macOS target platforms, and the `resources` directory for other target platforms. The
      * resources directory can be referenced in the packaged app via the
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -335,9 +335,9 @@ declare namespace electronPackager {
     extendInfo?: string | { [property: string]: any }; // eslint-disable-line @typescript-eslint/no-explicit-any
     /**
      * When the value is a string, specifies the filename of a `plist` file. Its contents are merged
-     * into all the Helper app's `Info.plist` files.
+     * into all the Helper apps' `Info.plist` files.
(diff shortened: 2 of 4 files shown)


src/mac.js
```diff
@@ -237,6 +251,9 @@ class MacApp extends App {
 
     if (this.usageDescription) {
       for (const [type, description] of Object.entries(this.usageDescription)) {
+        for (const plistKey of plistsToUpdate) {
+          this[plistKey][`NS${type}UsageDescription`] = description
+        }
```
ckerr, line 256: _nit_ `NS${type}UsageDescription` is built length+1 times; could be moved outside of the inner loop and built just once

src/mac.js
```diff
@@ -189,21 +189,31 @@ class MacApp extends App {
...
+    }
+
+    this.helperPlist = this.updateHelperPlist(this.helperPlist)
     for (const [plistKey, ...suffixArgs] of updateIfExists) {
       if (!this[plistKey]) continue
       this[plistKey] = this.updateHelperPlist(this[plistKey], ...suffixArgs)
     }
 
+    // Some properties need to go on all helpers as well, version, usage info, etc.
+    const plistsToUpdate = updateIfExists.filter(([key]) => !!this[key]).map(([key]) => key).concat(['appPlist', 'helperPlist'])
```
ckerr, line 215: This would be easier to read if there was a newline between each step

src/mac.js
```diff
@@ -189,21 +189,31 @@ class MacApp extends App {
...
+
     const updateIfExists = [
       ['helperRendererPlist', '(Renderer)', true],
       ['helperPluginPlist', '(Plugin)', true],
       ['helperGPUPlist', '(GPU)', true],
       ['helperEHPlist', 'EH'],
       ['helperNPPlist', 'NP']
     ]
+
+    for (const [plistKey] of [...updateIfExists, ['helperPlist']]) {
```
ckerr, line 203: The first arg of each array in `updateIfExists` is used as a key. Maybe this variable should be an object instead of an array of arrays e.g.

```
    const updateIfExists = { 
      'helperRendererPlist': ['(Renderer)', true],
      'helperPluginPlist': ['(Plugin)', true],
      'helperGPUPlist': ['(GPU)', true],
      'helperEHPlist': ['EH'],
      'helperNPPlist': ['NP']
    }
```

Not a big deal but it would make the code that uses it a little easier to read

src/mac.js
```diff
@@ -212,11 +222,15 @@ class MacApp extends App {
     }
 
     if (this.appVersion) {
-      this.appPlist.CFBundleShortVersionString = this.appPlist.CFBundleVersion = '' + this.appVersion
+      for (const plistKey of plistsToUpdate) {
+        this[plistKey].CFBundleShortVersionString = this[plistKey].CFBundleVersion = '' + this.appVersion
```
ckerr, line 226: `'' + this.appVersion` could be moved outside of the loop to ensure it's only built once. Same for buildVersion below

ckerr (commented):
Some readability suggestions inline but nothing major. Overall looks reasonable.

malept (commented):
I'll take a closer look once Charles's comments are addressed.

src/mac.js
MarshallOfSound: It's used as a key on a different object, I think this would be harder to read because you'd have to pull the keys and values separately via `entries()` or with two calls two `keys()` and `values()`.  Also code hasn't been introduced here, in the spirit of minimal and related changes we probably shouldn't refactor here.

MarshallOfSound:
Updated as per PR comments @ckerr @malept

NEWS.md
```diff
@@ -14,6 +14,10 @@
 * Support for customizing Windows targets on darwin/arm64 (#1260)
 * Support for customizing Windows targets on WSL without Wine installed (#1260)
 
+### Added
+
+* `extendHelperInfo` option to allow extending helper app `Info.plist` files
+
```
malept, line 20: Needs to go above the 15.3.0 release notes (also reference this PR while you're at it)

src/index.d.ts
```diff
@@ -333,6 +333,19 @@ declare namespace electronPackager {
      * @category macOS
      */
     extendInfo?: string | { [property: string]: any }; // eslint-disable-line @typescript-eslint/no-explicit-any
+    /**
+     * When the value is a string, specifies the filename of a `plist` file. Its contents are merged
+     * into all the Helper app's `Info.plist` files.
```
malept, line 338: Nit(grammar): pretty sure it's a plural possessive because it's multiple helper apps.

```suggestion
     * into all the Helper apps' `Info.plist` files.
```

src/mac.js
```diff
@@ -132,16 +132,16 @@ class MacApp extends App {
     return this.updatePlist(base, `${this.appName} ${helperSuffix}`, identifier, name)
   }
 
-  async extendAppPlist (propsOrFilename) {
+  async extendPlist (base, propsOrFilename) {
```
malept, line 135: Nit(naming): I want to be specific about this mostly because I see `baseDir` a lot in the tooling, and just a `base` variable might be confusing.
```suggestion
  async extendPlist (basePlist, propsOrFilename) {
```

src/mac.js
```diff
@@ -132,16 +132,16 @@ class MacApp extends App {
...
     if (!propsOrFilename) {
       return Promise.resolve()
     }
 
     if (typeof propsOrFilename === 'string') {
       const plist = await this.loadPlist(propsOrFilename)
-      return Object.assign(this.appPlist, plist)
+      return Object.assign(base, plist)
```
malept, line 142:
```suggestion
      return Object.assign(basePlist, plist)
```

src/mac.js
```diff
@@ -132,16 +132,16 @@ class MacApp extends App {
...
     } else {
-      return Object.assign(this.appPlist, propsOrFilename)
+      return Object.assign(base, propsOrFilename)
```
malept, line 144:
```suggestion
      return Object.assign(basePlist, propsOrFilename)
```

src/mac.js
```diff
@@ -212,11 +225,16 @@ class MacApp extends App {
...
+      for (const plistKey of plistsToUpdate) {
+        this[plistKey].CFBundleShortVersionString = this[plistKey].CFBundleVersion = appVersionString
+      }
     }
 
     if (this.buildVersion) {
-      this.appPlist.CFBundleVersion = '' + this.buildVersion
+      for (const plistKey of plistsToUpdate) {
+        this[plistKey].CFBundleVersion = '' + this.buildVersion
+      }
```
malept, line 237: For consistency:
```suggestion
      const buildVersionString = '' + this.buildVersion
      for (const plistKey of plistsToUpdate) {
        this[plistKey].CFBundleVersion = buildVersionString
      }
```

malept (changes requested):
This looks good conceptually. Just some cleanup needed, then we should be good to merge.

malept (approved):

LarryKlugerDS:
Hi @MarshallOfSound , This looks great!
Can it be used to set the `info.plist` for an app's private protocol? (Set via [app.setAsDefaultProtocolClient](https://www.electronjs.org/docs/api/app#appsetasdefaultprotocolclientprotocol-path-args))

Can you supply instructions on how to do so? 
Many thanks, Larry

PR #1281, opened 2021-09-15
fix: remove ElectronAsarIntegrity when asar is disabled

MarshallOfSound:
When `asar` is disabled we should remove the `ElectronAsarIntegrity` field from the `Info.plist`


malept (dismissed):
Looks good, add a test?

malept:
I guess fix the tests too :sweat_smile:

erickzhao (approved):

PR #1352, opened 2022-04-07
debug notarize options

devsibwarra:
* [X] I have read the [contribution documentation](https://github.com/electron/electron-packager/blob/main/CONTRIBUTING.md) for this project.
* [X] I agree to follow the [code of conduct](https://github.com/electron/electron/blob/main/CODE_OF_CONDUCT.md) that this project follows, as appropriate.
* [ ] The changes are appropriately documented (if applicable).
* [ ] The changes have sufficient test coverage (if applicable).
* [ ] The testsuite passes successfully on my local machine (if applicable).

**Summarize your changes:**
Add debug output for electron-notarize options

### DIFF
--- a/src/mac.js
+++ b/src/mac.js
@@ -366,6 +366,7 @@ class MacApp extends App {
         this.opts.quiet
       )
       if (notarizeOpts) {
+        debug(`Running electron-notarize with the options ${JSON.stringify(notarizeOpts)}`)
         return notarize(notarizeOpts)
       }
     }


MarshallOfSound (changes requested):
These options can contain credentials, we should avoid logging them

devsibwarra:
> These options can contain credentials, we should avoid logging them

Okay. I assumed it wasn't a concern because the `Packager Options` debug output [here](https://github.com/electron/electron-packager/blob/a0ef38f08a577a3a0491fd5db3eb1f55d2e44d2b/src/index.js#L165) could contain the same credentials

PR #1361, opened 2022-04-27
fix: package should not log info on --quiet flag

### LINKED ISSUE #1333: Prints `Wrote new app to ...` when `--quiet` is provided
### Preflight Checklist

### Issue Details

* **Electron Packager Version:** 15.4.0
* **Electron Version:** 17.0.1
* **Operating System:** Windows 10 21H2

### Expected Behavior
electron-packager should not print to the console when the `--quiet` option is provided.

### Actual Behavior
electron-packager prints `Wrote new app to ...` when the `--quiet` option is provided.

### To Reproduce
```bash
npx electron-packager . --quiet --out dist --overwrite
```

erickzhao:
Fixes #1333

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const { hostInfo, warning } = require('./common')
+const { common, hostInfo, warning } = require('./common')
 const fs = require('fs-extra')
 const { initializeProxy } = require('@electron/get')
 const packager = require('..')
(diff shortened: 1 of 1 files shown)


ckerr (approved):
:+1:

ioniall:
It appears this breaks the package. An error appears `Cannot read properties of undefined (reading 'info')` when using electron-packager to package a build. The `import { common...` should probably have been `import { info...` and the matching `common.info(` logs be `info(` logs.

PR #1391, opened 2022-08-02
fix: ignore node_gyp_bins if it exists

VerteDinde:
* [x] The testsuite passes successfully on my local machine (if applicable). (Passes on MacOS Monterey x64 & arm64)

Addresses https://github.com/electron/electron-rebuild/issues/1024

In node-gyp v9.0.0, node-gyp adds a python symlink for older versions of python. This symlink is added to a `${module_name}/build/node_gyp_bins/` directory within the build directory, which electron-rebuild then tries to rebuild.

This PR ignores the `node_gyp_bins` directory, and adds tests for making sure that the directory is not added to the packaged app.


test/darwin.js
```diff
@@ -482,7 +482,7 @@ if (!(process.env.CI && process.platform === 'win32')) {
     t.deepEqual(plistObj.ElectronAsarIntegrity, {
       'Resources/app.asar': {
         algorithm: 'SHA256',
-        hash: '27f2dba4273f6c119000ec7059c27d86e27306d5dbbb83cfdfb862d92c679574'
+        hash: '2ef4c03f6fc8adf03f782b40fddcfcdd26d1f3f3669ecedffeea5de5d28897c3'
```
VerteDinde, line 485: I'm oretty sure that this hash changed because I updated the `basic` app that `baseOpts` is passing into the test. Tested this locally on an x64 machine, and the hash is consistently returning this value both locally and in CI. However, if this doesn't seem to pass the sniff test, let me know, happy to take another look.

malept:
Hmmm. Separate PR to fix the xcode version in use, so the tests pass?

VerteDinde:
@malept Yeah, hilariously that image failure is new as of this afternoon's rerun 😄 I'll fix that in another PR and then rebase

test/darwin.js
malept: > I'm oretty sure that this hash changed because I updated the basic app that baseOpts is passing into the test.

Might want to add a comment to this effect for future folks (and also how to find out the new value).

malept (approved):
Aside from a maintenance comment, this seems fine.

erickzhao (approved):

PR #1425, opened 2022-10-04
Fix node version check for legacy projects

Duelsik:
(maybe worth noting: the testsuite had weird fail on the first run when downloading electrons)

**Summarize your changes:**

Comparison between tables worked by accident when node version wasn't starting with "1" when toStringed.
I am aware that 8.7.x is superlegacy branch, but I've found a project that I would like to support and it depends on this version.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -2,8 +2,9 @@
 
 'use strict'
 
-var nodeVersionInfo = process.versions.node.split('.').map(function (n) { return Number(n) })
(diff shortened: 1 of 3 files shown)


Duelsik:
Not sure what to do with this CircleCI config - pass an empty step that will just pass?

Duelsik:
Thanks for your time, I will look into how much work it would take to update this project to modern standard.

PR #1439, opened 2022-11-01
fix: allow packaging twice simultaneously

MarshallOfSound:
Packager didn't use unique staging / temp directories so running packager twice with similar targets would cause issues.


VerteDinde (approved):

VerteDinde (commented):
~Failing CI looks like a flake~ I am wrong

PR #1441, opened 2022-11-04
chore: update galactus

MarshallOfSound:
This updates galactus to the latest `^1.0.0` release which contains 0 code changes but bumps some dependencies resulting in a new minimum node version of 12 (which is the same as packager currently).

This PR also adds a lockfile because it's unsafe for this repository to not have one 🙃


erickzhao (approved):
seems legit. 👍 lockfile

PR #1448, opened 2022-11-20
feat: add support to @electron/fuses

erikian:

**Summarize your changes:**

This PR brings [@electron/fuses](https://github.com/electron/fuses) support to Packager. I thought it would be better to implement this feature in Packager rather than Forge (see https://github.com/electron/forge/issues/3076) because, well, Forge gets automatic support to it and users running Packager directly also get to use this feature.

I didn't add any tests because testing this feature would require actually packaging an app, and looks like no other test does it ATM, so I decided to check if it's ok first.

Closes https://github.com/electron/forge/issues/3076.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
   ],
   "dependencies": {
     "@electron/asar": "^3.2.1",
+    "@electron/fuses": "^1.6.1",
     "@electron/get": "^2.0.0",
     "@electron/notarize": "^1.2.3",
     "@electron/osx-sign": "^1.0.1",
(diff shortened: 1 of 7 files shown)


erikian:
CircleCI had me set up a project for my fork and now looks like it's complaining about not finding a `yarn.lock`. I'm not familiar with CircleCI, so I'm not sure if it's expected, the jobs just worked the last time I sent a PR to Packager.

image

(On a side note, this project should really have a `yarn.lock` on Git, right? 😀)

PR #1478, opened 2023-02-15
Add basename option

Cefold:

**Summarize your changes:**
Add basename option to `generateFinalBasename`, that we could avoid [this error](https://github.com/nodejs/node-gyp/issues/65) if the appname has space.

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -12,7 +12,7 @@ function sanitizeAppName (name) {
 }
 
 function generateFinalBasename (opts) {
-  return `${sanitizeAppName(opts.name)}-${opts.platform}-${opts.arch}`
+  return `${sanitizeAppName(opts.basename || opts.name)}-${opts.platform}-${opts.arch}`
 }
 
 function generateFinalPath (opts) {
--- a/src/copy-filter.js
+++ b/src/copy-filter.js
@@ -40,6 +40,7 @@ function generateIgnoredOutDirs (opts) {
       for (const arch of archs) {
         const basenameOpts = {
           arch: arch,
+          basename: opts.basename,
           name: opts.name,
           platform: platform
         }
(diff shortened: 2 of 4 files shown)


malept (commented):
I think there are better solutions for this problem than adding **yet another** confusing option to Packager, but I'll let other Ecosystem WG members weigh in.

Cefold:
> I think there are better solutions for this problem than adding **yet another** confusing option to Packager, but I'll let other Ecosystem WG members weigh in.

We could also try this: `${sanitizeAppName(opts.name.replace(' ', ''))}-${opts.platform}-${opts.arch}`

malept:
Regardless of the implementation,  it'd be best if you added tests which showed what the intended behavior is supposed to be.

Cefold:
Any recommendation from WG-ecosystem?

felixrieseberg:
I _just_ came across this problem, too. This solution here would fix my problem.

The real problem is that `electron-packager` lets you disable the use of `tmpdir` but doesn't let you control what the folder is named it'll use to package. I'd love to override the folder name.

@malept I think I'm generally in favor of merging it in.

malept:
>  have it use name over productName for the folder naming.

I'd be more inclined to do this than add another flag, but I currently lack the context of what might break if we did that.

felixrieseberg:
In theory, the internal folder structure is an implementation detail and not part of the API, so we _could_ change it in a minor version.

felixrieseberg:
I'd love to fix this problem! I tried a few things here and I believe that changing how we generate the name is the best thing we can do, because _either_ solution will require a change in `electron-forge`. 

I'm closing this PR here. @Cefold I am so sorry about how long it took for us to actually act here.

PR #1590, opened 2023-11-02
build!: update minimum Node.js version from `14.17.5` to `16.0.0`

erikian:
As in title.


VerteDinde (approved):

dsanders11 (changes requested):
So that `semantic-release` picks up the breaking change and versions correctly, fixup the commit to add "BREAKING CHANGE: foo" in the commit body.

dsanders11 (approved):

.circleci/config.yml
```diff
@@ -61,7 +61,6 @@ workflows:
                 - 20.2.0
                 - 18.14.0
                 - 16.19.0
```
BlackHole1, line 63: Perhaps we should conduct tests based on version 16.0.0.

```suggestion
                - 16.0.0
```

BlackHole1 (changes requested):

BlackHole1 (approved):

PR #1593, opened 2023-11-03
refactor: TypeScript rewrite

erikian:

**Summarize your changes:**


src/prune.ts
```diff
@@ -1,5 +1,5 @@
 import { normalizePath, warning } from './common';
-import galactus from 'galactus';
+import { DestroyerOfModules, DepType, Module, ModuleMap } from 'galactus';
```
felixrieseberg, line 2: The names will always be funny to me

src/hooks.ts
```diff
@@ -1,21 +1,25 @@
...
-    const args = Array.prototype.splice.call(arguments, 0, arguments.length - 1);
-    const done = arguments[arguments.length - 1];
-    for (const hook of hooks) {
-      await hook.apply(this, args);
-    }
-
-    return done(); // eslint-disable-line promise/no-callback-in-promise
-  };
-}
+/**
```
felixrieseberg, line 12: You used to be able to run `const { serialHooks } = require('electron-packager/hooks')`. I'm not sure if we actually need that though, people could just implement it themselves?

felixrieseberg (approved):
I gave this whole PR one solid read - and even though it's massive and I think I should have found _one_ place to provide constructive feedback, this seems mostly pretty good to me.

The one thing we should make double-and-triple sure before merging is that nobody can actually rely on `electron-packager/hooks` - and if they could, that we maybe just keep it and cut it in a later PR.

src/hooks.ts
malept: Getting rid of it would require a major version bump. I literally added it because someone complained about the behavior breaking, so it's probably used in at least one app. It's sounding like we're going to try to bring this PR in as a minor version bump, so we should probably keep it for now (and label it as deprecated?)

src/common.ts
```diff
@@ -0,0 +1,127 @@
...
+
+export function ensureArray<T>(value: T | T[]): T[] {
+  return Array.isArray(value) ? value : [value];
+}
+
+export function isPlatformMac(platform: ComboOptions['platform']) {
+  return platform === 'darwin' || platform === 'mas';
+}
+
+// @TODO(erikian): the only uses of this function are in its own test - remove it?
```
malept, line 66: I left it there because it might be useful in the future. I don't have a strong opinion on whether to keep it currently.

.circleci/config.yml
```diff
@@ -21,7 +21,9 @@ jobs:
       - node/test:
           post-node-js-install-steps:
             - run: test/ci/before_install.sh
+          override-ci-command: yarn install --frozen-lockfile --ignore-engines
```
malept, line 24: question: what dependency requires us to add `--ignore-engines`?

.gitignore
```diff
@@ -13,3 +13,5 @@ typedoc
 .DS_Store
 .nyc_output
 npm-debug.log
+dist/
+.idea
```
malept, line 17: Nit: arguably, editor-specific ignores should be in the user's global git ignore.

package.json
```diff
@@ -2,8 +2,8 @@
   "name": "electron-packager",
   "version": "17.1.2",
   "description": "Customize and package your Electron app with OS-specific bundles (.app, .exe, etc.) via JS or CLI",
-  "main": "src/index.js",
-  "types": "src/index.d.ts",
+  "main": "dist/index.js",
+  "types": "dist/types.d.ts",
```
malept, line 6: thought: if we're going to be switching to TypeScript anyway, perhaps in a follow-up PR we should add ESM support.

package.json
```diff
@@ -2,8 +2,8 @@
...
   "bin": {
     "electron-packager": "bin/electron-packager.js"
```
malept, line 8: issue: I'm pretty sure you'll need to update this file. (Although I don't know if I'd necessarily convert it to TypeScript, since it's basically a Node.js-version-checking shim for `src/cli`.)

.eslintrc.js
```diff
@@ -0,0 +1,46 @@
...
+  'plugin:@typescript-eslint/eslint-recommended',
+  'plugin:@typescript-eslint/recommended',
+  'plugin:import/typescript',
+);
+
+eslintConfig.rules['comma-dangle'] = ['error', 'only-multiline'];
+eslintConfig.rules.semi = ['error', 'always'];
+eslintConfig.rules['space-before-function-paren'] = ['error', 'never'];
+
+module.exports = eslintConfig;
```
malept, line 46: suggestion: this can definitely be merged with the `eslintConfig` above. In fact, this can probably be a JSON file instead of a JS file (although I personally prefer it in `package.json` because I hate all of the dotfiles in the root directory).

package.json
erikian: We're still shipping the `bin` directory to npm, so the path here is fine — I did forgot to update a `require` in that file from `src` to `dist` though 😅

That being said, do we really need this file? Like, if the user has e.g. Node 10 and we require Node 16, the package manager should prevent `packager` from being installed (unless they're using `--ignore-engines`, but then the user should be aware that things can break)

.circleci/config.yml
erikian: In my case, `ts-api-utils` via `@typescript-eslint#eslint-plugin`, see [this failed run](https://app.circleci.com/pipelines/github/electron/electron-packager/1054/workflows/da09ed73-6f94-4b00-83cc-f3eaf623fe44/jobs/5885?invite=true#step-113-204_120). I actually copied this line from the [CFA releases PR](https://github.com/electron/electron-packager/pull/1566/commits/332ae93f798d97b0612599fef974ac8f2fd357a6#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R25), I imagine it was included there for the same reason.

package.json
erikian: Sounds great!

.gitignore
erikian: Agreed, this slipped in by accident

package.json
malept: You'd be surprised how many issues were filed back in the day about why there were SyntaxErrors even though the minimum engine has been specified for a very long time.

src/hooks.ts
erikian: Did some digging and found the use case for this: https://github.com/electron/electron-packager/commit/a2f3a89e55ac2da639de325856a6924a80c896b1#diff-9eddf4dc51bf6a0125ca7fb094468ad284112270385c41cebce4f8f0a29620abR44-R68. The contents of `docs/api.md` were eventually replaced by a link to https://electron.github.io/electron-packager, so we lost the only mention of this function we had in the project.

That makes me thing we should keep it and just add a comment with that missing docs. We could also make it possible to import it from the top-level package (`import { serialHooks } from 'electron-packager'`) and deprecate / add a deprecation notice for `electron-packager/hooks` (but not for the function itself) if that makes sense.

.eslintrc.js
erikian: A lot of our repos have dotfiles instead of keeping multiple configs in `package.json`, so the idea here was to keep Packager consistent with other projects. A nice thing about using .js files for configuration is they allow [adding comments](https://github.com/electron/fiddle/blob/main/.eslintrc.js), so IMO that justifies having a bunch of dotfiles.

dsanders11:
Something to watch out for, there's at least one usage in Forge that directly reaches into `src/`: https://github.com/electron/forge/blob/fd00d9abf1dfc6b34deadfea62ea72d4589105c8/packages/api/core/src/util/parse-archs.ts#L5

package.json
malept: I happened to find a relatively new instance of this with Electron Forge in the [Discord server](https://discord.com/channels/745037351163527189/1175850453637009518): the user was using Node.js 12 when Forge requires Node.js 16, and Forge definitely has engine constraints in its various package.json files.

src/hooks.ts
erikian: https://github.com/electron/packager/pull/1593/commits/47e4006c666e72aea8347bc98c19db38bb826d28

package.json
erikian: Yeah, now I see why the validation is there, we shouldn't trust `engines` at all. Adding a proper validation to the ~40 packages in the Forge monorepo is gonna be fun 😟

malept: I would just add it to the two CLI-facing packages: `create-electron-app` and `@electron-forge/cli`.

erikian:
> Something to watch out for, there's at least one usage in Forge that directly reaches into `src/`: https://github.com/electron/forge/blob/fd00d9abf1dfc6b34deadfea62ea72d4589105c8/packages/api/core/src/util/parse-archs.ts#L5

early draft so I don't forget it: https://github.com/electron/forge/pull/3419/commits/c9da7148961a9a3cb2aee119bec89b5e533d2dd3

src/common.ts
erikian: Let's remove it, it hasn't been used for a good 6 years and we can always recover it from the Git history if necessary - https://github.com/electron/packager/pull/1593/commits/32af02c78e2f62a741b49b30502b3acbd38fbb96

erikian:
My n=1 experiment says removing [this hack](https://github.com/electron/packager/blob/4911693b52166d8016452d421894abd3a94ac2fc/src/cli.js#L106-L111), originally introduced in #381, doesn't result in a catastrophic failure. The PR only mentions some truncated text when displaying the usage info, so we _should_ be good 🤞🏼

dsanders11:
@erikian, couple of conflicts due to other PRs landing.

PR #1609, opened 2023-11-22
feat: Add @electron/windows-sign

felixrieseberg:

**Summarize your changes:**

This PR adds @electron/windows-sign, enabling this module to codesign Windows binary files. Both logic and configuration are analogous to @electron/osx-sign -- on purpose, to keep the API simple and easy to use.

### DIFF
--- a/src/win32.js
+++ b/src/win32.js
@@ -3,6 +3,7 @@
 const debug = require('debug')('electron-packager')
 const path = require('path')
 const { WrapperError } = require('cross-spawn-windows-exe')
+const { sign } = require('@electron/windows-sign')
 
 const App = require('./platform')
 const common = require('./common')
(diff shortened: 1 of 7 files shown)


src/win32.js
```diff
@@ -98,15 +99,53 @@ class WindowsApp extends App {
...
+  if (typeof properties === 'object') {
+    result = { ...properties, appDirectory }
+  }
+
+  // A little bit of convenience
+  if (windowsMetaData) {
+    if (windowsMetaData.FileDescription) {
+      result.description = windowsMetaData.FileDescription
+    }
+  }
```
MarshallOfSound, line 144: Should this really supercede the user provided value? Probably should add a `!description` check to avoid overwriting.  Also collapse these if statements

```suggestion
  // A little bit of convenience
  if (windowsMetaData && windowsMetaData.FileDescription && !result.description) {
    result.description = windowsMetaData.FileDescription
  }
```

src/win32.js
```diff
@@ -98,15 +99,53 @@ class WindowsApp extends App {
...
+      }
+    }
+  }
+
   async create () {
     await this.initialize()
     await this.renameElectron()
     await this.copyExtraResources()
     await this.runRcedit()
+    await this.signAppIfSpecified()
```
MarshallOfSound, line 127: Is this definitely run after the `afterCopy` hook? Just double checking this is run _after_ any potential mutations to the exe.  E.g. via fuses

felixrieseberg: Yep! The only thing that happens after is moving the package to its final destination and the `afterComplete` hook. For future convenience, this is also exactly when we sign & notarize on macOS.

erikian (approved):

L11R:
@felixrieseberg how exactly I should use it? I tried object with just `signWithParams` (we use cert from system storage), but it throws path-related error. I tried to specify `appDirectory`, it goes further but there are no binary at this moment yet (they are in some packager temporary dir).

PR #1657, opened 2024-01-10
fix: copyTemplate failed if tmpdir set to false

ianho:

**Summarize your changes:**

if tmpdir set to false, `fs-extra copy` api will throw a error that `Cannot copy 'xxx' to a subdirectory of itself, 'xxx/out/xxx'.`

so we cant call `fs.copy` in this case, traverse first-level directories to call `fs.copy` instead

### DIFF
--- a/src/platform.ts
+++ b/src/platform.ts
@@ -153,10 +153,27 @@ export class App {
   async copyTemplate() {
     await promisifyHooks(this.opts.beforeCopy, this.hookArgsWithOriginalResourcesAppDir);
 
-    await fs.copy(this.opts.dir, this.originalResourcesAppDir, {
-      filter: userPathFilter(this.opts),
-      dereference: this.opts.derefSymlinks,
-    });
+    if (this.opts.tmpdir === false) {
+      const filter = userPathFilter(this.opts);
+      const items = await fs.readdir(this.opts.dir);
+      await Promise.all(items.map(async item => {
+        const srcItem = path.join(this.opts.dir, item);
+        const destItem = path.join(this.originalResourcesAppDir, item);
+
+        const include = await filter(srcItem, destItem);
+        if (!include) return;
+
+        return fs.copy(srcItem, destItem, {
+          filter,
+          dereference: this.opts.derefSymlinks ?? true,
+        });
+      }));
+    } else {
+      await fs.copy(this.opts.dir, this.originalResourcesAppDir, {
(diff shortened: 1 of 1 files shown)


src/platform.ts
```diff
@@ -153,10 +153,27 @@ export class App {
...
+      const items = await fs.readdir(this.opts.dir);
+      await Promise.all(items.map(async item => {
+        const srcItem = path.join(this.opts.dir, item);
+        const destItem = path.join(this.originalResourcesAppDir, item);
+
+        const include = await filter(srcItem, destItem);
+        if (!include) return;
+
+        return fs.copy(srcItem, destItem, {
+          filter,
```
BlackHole1, line 167:
```suggestion
```

Since the judgment has already been made before, there is no need to let fs.copy execute the judgment again.

BlackHole1 (changes requested):
Thank you for your PR!
It would be better if you could explain in detail in the PR why this issue occurred. From the existing PR description and code changes, it is difficult for me to understand the reason behind them. I only understood the reason after reviewing the source code of [fs.copy](https://github.com/jprichardson/node-fs-extra/blob/acf55855580db460fd5532b22d73c4543798cd7d/lib/copy/copy.js#L27-L33) (because the `condition` is checked before the `filter`).

BlackHole1:
PTAL @malept @erikian

src/platform.ts
```diff
@@ -153,10 +153,27 @@ export class App {
   async copyTemplate() {
     await promisifyHooks(this.opts.beforeCopy, this.hookArgsWithOriginalResourcesAppDir);
 
-    await fs.copy(this.opts.dir, this.originalResourcesAppDir, {
-      filter: userPathFilter(this.opts),
-      dereference: this.opts.derefSymlinks,
-    });
+    if (this.opts.tmpdir === false) {
```
erickzhao, line 156: Hey @ianho, we discussed this in an Ecosystem WG meeting and I think this PR is the best solution for https://github.com/electron/forge/issues/3475 as well since it resolves the problem upstream.

Two blockers here:

* Can we add a comment explaining why this is necessary about line 156?
* We should add a test for this code so it doesn't regress in the future.

erickzhao (changes requested):

MarshallOfSound:
Superseded by https://github.com/electron/packager/pull/1927

PR #1696, opened 2024-03-25
feat: remove dependency on rcedit to allow x-platform exe resource modding

MarshallOfSound:
This re-implements _everythin_ that `rcedit` used to do with a native binary using a purely JS implementation of resource editing.

This is required to drop the package-on-platform requirements of the win32 target, it also will make it easier to add asar integrity windows manifest values in the future.

### DIFF
--- a/src/resedit.ts
+++ b/src/resedit.ts
@@ -35,7 +35,7 @@ function parseVersionString(str: string): ParsedVersionNumerics {
 // Ref: Link MSDN
 const RT_MANIFEST_TYPE = 24;
 
-export async function rcedit(exePath: string, options: ExeMetadata) {
+export async function resedit(exePath: string, options: ExeMetadata) {
   const resedit = await loadResEdit();
 
   const exeData = await fs.readFile(exePath);
--- a/src/resedit.ts
+++ b/src/resedit.ts
@@ -32,7 +32,7 @@ function parseVersionString(str: string): ParsedVersionNumerics {
   }) as ParsedVersionNumerics;
 }
 
-// Ref: Link MSDN
+// Ref: https://learn.microsoft.com/en-us/windows/win32/menurc/resource-types
 const RT_MANIFEST_TYPE = 24;
 
 export async function resedit(exePath: string, options: ExeMetadata) {
--- a/src/win32.ts
+++ b/src/win32.ts
@@ -35,16 +35,6 @@ export class WindowsApp extends App {
       productName: this.opts.name,
       win32Metadata,
     };
-
-    // TODO
-    // const manifestProperties = ['application-manifest', 'requested-execution-level'];
-    // for (const manifestProperty of manifestProperties) {
-    //   if (win32metadata[manifestProperty as keyof typeof win32metadata]) {
-    //     // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-    //     // @ts-ignore
-    //     rcOpts[manifestProperty] = win32metadata[manifestProperty];
-    //   }
-    // }
   }
 
   async getIconPath() {
--- a/src/win32.ts
+++ b/src/win32.ts
@@ -32,7 +32,7 @@ export class WindowsApp extends App {
       productVersion: this.opts.appVersion,
       fileVersion: this.opts.buildVersion || this.opts.appVersion,
       legalCopyright: this.opts.appCopyright,
-      productName: this.opts.name,
+      productName: this.opts.win32metadata?.ProductName || this.opts.name,
       win32Metadata,
     };
   }
@@ -55,16 +55,16 @@ export class WindowsApp extends App {
       return Promise.resolve();
     }
 
-    const rcOpts = this.generateReseditOptionsSansIcon();
+    const resOpts = this.generateReseditOptionsSansIcon();
 
     // Icon might be omitted or only exist in one OS's format, so skip it if normalizeExt reports an error
     const icon = await this.getIconPath();
     if (icon) {
-      rcOpts.iconPath = icon;
+      resOpts.iconPath = icon;
     }
 
-    debug(`Running resedit with the options ${JSON.stringify(rcOpts)}`);
-    await resedit(this.electronBinaryPath, rcOpts);
+    debug(`Running resedit with the options ${JSON.stringify(resOpts)}`);
+    await resedit(this.electronBinaryPath, resOpts);
   }
 
   async signAppIfSpecified() {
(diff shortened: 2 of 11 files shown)


src/resedit.ts
```diff
@@ -0,0 +1,118 @@
...
+  return parts.map((part) => {
+    const parsed = parseInt(part, 10);
+    if (isNaN(parsed)) {
+      throw new Error(`Incorrectly formatted version string: "${str}". Component "${part}" could not be parsed as an integer`);
+    }
+    return parsed;
+  }) as ParsedVersionNumerics;
+}
+
+// Ref: Link MSDN
```
felixrieseberg, line 35: Missing link?

src/win32.ts
```diff
@@ -31,39 +19,32 @@ export class WindowsApp extends App {
...
+    return {
+      productVersion: this.opts.appVersion,
+      fileVersion: this.opts.buildVersion || this.opts.appVersion,
+      legalCopyright: this.opts.appCopyright,
+      productName: this.opts.name,
+      win32Metadata,
+    };
 
-    return rcOpts;
+    // TODO
```
felixrieseberg, line 39: This might be a breaking change

felixrieseberg (commented):
I took a gander through the resedit source code (https://github.com/jet2jet/resedit-js/blob/main/package.json) and it seems really well done, so I'm generally supportive. 

I found some code that looks a little WIP but I think this is overall net-good.

src/win32.ts
MarshallOfSound: Oh I implemented support for these, just didn't delete the commented out code 😅 Not breaking

MarshallOfSound:
> I took a gander through the resedit source code (https://github.com/jet2jet/resedit-js/blob/main/package.json) and it seems really well done, so I'm generally supportive.

Yeah for context I started out implementing this myself, then hit an edge case I didn't fully understand. In my search for the answer I found an impl that was very similar to what I was writing from scratch so decided to use that impl instead.  It's 0 dependencies, and works quite well.

src/resedit.ts
erikian, line 1: The value for the `win32metadata.ProductName` option is not being written to the .exe anymore, which I believe would be a breaking change. Here's a sample `package.json`:

```json
{
  "name": "resedit-test-app",
  "version": "1.0.0",
  "main": "src/index.js",
  "scripts": {
    "package": "electron-packager . --win32metadata.ProductName=\"Product Name from Packager Options\" --overwrite"
  },
  "devDependencies": {
    "@electron/packager": "file:C:\\path\\to\\local-electron-packager",
    "electron": "29.1.5"
  }
}
```

image

src/win32.ts
```diff
@@ -74,30 +45,26 @@ export class WindowsApp extends App {
     return this.normalizeIconExtension('.ico');
   }
 
-  needsRcedit() {
-    return Boolean(this.opts.icon || this.opts.win32metadata || this.opts.appCopyright || this.opts.appVersion || this.opts.buildVersion);
+  needsResedit() {
+    return Boolean(this.opts.icon || this.opts.win32metadata || this.opts.appCopyright || this.opts.appVersion || this.opts.buildVersion || this.opts.name);
```
erikian, line 49: Looks like this was broken before, but I believe this might _technically_ be a breaking change in the following scenario:

- the user-provided Packager config does not specify any of the options we're checking for here
- the app's `package.json` does not specify the `version` and `author` fields

In that case, Packager is currently not editing the .exe metadata, and with this change it would use the `name` field from the Packager options (or from the apps's `package.json` `name` / `productName` fields) for the file description, product name and original filename fields. That makes sense, but should we be concerned about someone relying on these fields being just `Electron` / `electron.exe` even if they're passing different values or nah?

Sample `package.json`:

```json
{
  "name": "resedit-test-app",
  "//productName": "Product Name from packageJson",
  "main": "src/index.js",
  "scripts": {
    "package": "electron-packager . --overwrite"
  },
  "devDependencies": {
    "@electron/packager": "file:C:\\path\\to\\local-electron-packager",
    "electron": "29.1.5"
  }
}
```

test/win32.js
```diff
@@ -4,9 +4,11 @@ const config = require('./config.json')
 const { packager } = require('../dist')
 const path = require('path')
 const test = require('ava')
+const _ = require('lodash')
```
erikian, line 7: We don't have `lodash` explicitly listed as a dev dep right now, should we add it?

src/resedit.ts
MarshallOfSound: This was never supported per docs, it worked due to a quirk in props spreading, not because it was intended afaik

erickzhao: I see this particular use case documented in `usage.txt`:

https://github.com/electron/packager/blob/6ffb8c3aa52f74178e7984253483533fb72e6949/usage.txt#L119-L122

src/win32.ts
```diff
@@ -1,22 +1,10 @@
 import path from 'path';
-import { WrapperError } from 'cross-spawn-windows-exe';
```
erickzhao, line 2: issue: i think we can fully remove `cross-spawn-windows-exe` from the deps as well

src/win32.ts
```diff
@@ -74,30 +45,26 @@ export class WindowsApp extends App {
...
-  async runRcedit() {
+  async runResedit() {
     /* istanbul ignore if */
-    if (!this.needsRcedit()) {
+    if (!this.needsResedit()) {
       return Promise.resolve();
     }
 
-    const rcOpts = this.generateRceditOptionsSansIcon();
+    const rcOpts = this.generateReseditOptionsSansIcon();
```
erickzhao, line 58: nit: rename this to `resOpts` or something for consistency

src/resedit.ts
MarshallOfSound: le sigh, it isn't in typescript or implementation. I'll add back support for it 😢

src/resedit.ts
```diff
@@ -0,0 +1,118 @@
...
+ * Parse a version string in the format a.b.c.d with each component being optional
+ * but if present must be an integer. Matches the impl in rcedit for compat
+ */
+function parseVersionString(str: string): ParsedVersionNumerics {
+  const parts = str.split('.');
+  if (parts.length === 0 || parts.length > 4) {
+    throw new Error(`Incorrectly formatted version string: "${str}". Should have at least one and at most four components`);
+  }
+  return parts.map((part) => {
+    const parsed = parseInt(part, 10);
```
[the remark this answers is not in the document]
MarshallOfSound: `Number` has these quirks too `Number('0xc') === 12`

erickzhao (approved):

PR #1697, opened 2024-03-26
feat: set asar integrity resource on windows packages

MarshallOfSound:
This is currently targeting my "remove rcedit" branch so that the diff is small and makes sense.

Pretty straight forward, sets the right asar integrity resource when packaging for windows.

### DIFF
--- a/src/win32.ts
+++ b/src/win32.ts
@@ -33,6 +33,7 @@ export class WindowsApp extends App {
       fileVersion: this.opts.buildVersion || this.opts.appVersion,
       legalCopyright: this.opts.appCopyright,
       productName: this.opts.win32metadata?.ProductName || this.opts.name,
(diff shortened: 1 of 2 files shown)


src/win32.ts
```diff
@@ -33,6 +33,7 @@ export class WindowsApp extends App {
       fileVersion: this.opts.buildVersion || this.opts.appVersion,
       legalCopyright: this.opts.appCopyright,
       productName: this.opts.name,
+      asarIntegrity: this.asarIntegrity,
```
erickzhao, line 36: https://github.com/electron/packager/blob/6ffb8c3aa52f74178e7984253483533fb72e6949/usage.txt#L119-L130 needs an update

erickzhao (approved):

PR #1698, opened 2024-03-28
build!: remove temporary shim

erikian:
This was added in #1641 to temporarily provide a missing file that Forge relied on until a new Forge version was released with an updated version of Packager, so we don't need this anymore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,4 +14,3 @@ typedoc
 .nyc_output
 npm-debug.log
 dist/
-src/targets.js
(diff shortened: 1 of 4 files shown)


dsanders11:
Isn't there still potential for issues here, if a user updates their version of `@electron/packager` without updating their version of Forge? Since Forge indicates it can use any minor version of v18 of `@electron/packager`, so someone on Forge v7.2.0 could still have a need for this.

It's probably an unlikely situation, but maybe we just leave the shim until the next breaking change in Packager?

erikian:
Yeah good point, this can wait for the next major release, [enough inadvertent breaking changes on v18](https://github.com/electron/packager/pull/1700) 😂

PR #1734, opened 2024-06-10
docs: typedoc improvements

erickzhao:
Cleanup PR based on learnings from the TypeDoc PRs.

* Direct use of `npx typedoc` rather than JavaScript API removes legacy `build_docs.js` from code. Their JavaScript API has minimal documentation, and most options are available via `typedoc.json` configuration anyways.
* Use of `externalSymbolLinkMappings` to show links to types available via external Electron packages (e.g. get and notarize).

Tag changes:
* Changed plaintext examples to use the `@example` tag.
* Fixed outstanding cases of legacy `[[link tag syntax]]` to use `{@link tag syntax}`.
* Used `@interface` to expand utility types whenever possible (e.g. instead of `Omit<TypeA, 'KeyB' | 'KeyC'>`, adding the `@interface` tag makes it so that all of `TypeA`'s properties are displayed in the API documentation excluding `KeyB` and `KeyC`.
* Used `@internal` and `excludeInternals` to hide types that should not be used by end users of Electron Packager.


package.json
```diff
@@ -69,7 +69,7 @@
     "nyc": "^15.0.0",
     "pkg-up": "^4.0.0",
     "sinon": "^17.0.0",
-    "typedoc": "^0.25.4",
+    "typedoc": "~0.25.4",
```
erickzhao, line 72: TypeDoc introduces breaking changes between SemVer minor versions (which is compliant since it has major version 0), so let's pin this to the same minor line.

typedoc.json
```diff
@@ -0,0 +1,14 @@
+{
+  "entryPoints": ["src/index.ts"],
+  "out": "./typedoc",
+  "excludeInternal": true,
+  "sort": ["kind", "required-first", "alphabetical"],
+  "externalSymbolLinkMappings": {
+    "@electron/notarize": {
+      "NotaryToolCredentials": "https://erickzhao.github.io/notarize/types/NotaryToolCredentials.html"
```
erickzhao, line 8: DONOTMERGE: let's merge in https://github.com/electron/notarize/pull/199 first!

package.json
malept: FWIW since it's pre-1.0, I think package managers do what you're asserting anyway.

test/ci/build_docs.js
```diff
@@ -1,41 +0,0 @@
...
-}
-
-const gitRevision = process.argv[2] || gitRevisionFromGitHubRef()
-if (gitRevision) {
-  if (/^[0-9a-f]+$/i.test(gitRevision)) {
-    config.gitRevision = gitRevision
-  } else if (gitRevision.startsWith('v')) {
-    config.includeVersion = true
-  }
-}
```
malept, line 30: thought: I'm pretty sure this is why I used the JS API instead of the CLI.

typedoc.json
```diff
@@ -0,0 +1,14 @@
+{
+  "entryPoints": ["src/index.ts"],
+  "out": "./typedoc",
+  "excludeInternal": true,
+  "sort": ["kind", "required-first", "alphabetical"],
+  "externalSymbolLinkMappings": {
```
malept, line 6: thought: oh, they finally added this?? Amazing

malept:
Do we also need to add explicit external link config for `osx-sign` and `windows-sign`? Plus any other dependency where we proxy the options that I'm forgetting.

felixrieseberg (approved):
Approving just so that you can merge when you want to!

PR #1742, opened 2024-06-26
chore: replace legacy `var` with `const`

erickzhao:
Now that `engines` requires Node.js 16

### DIFF
--- a/bin/electron-packager.js
+++ b/bin/electron-packager.js
@@ -2,15 +2,13 @@
 
 'use strict'
 
-/* eslint-disable no-var */
-// WHY: not consts so that this file can load in Node < 4.0
-var packageJSON = require('../package.json')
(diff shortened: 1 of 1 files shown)


BlackHole1 (approved):

bin/electron-packager.js
erikian, line 1: I planned to do this on #1593 but @malept [pointed out](https://github.com/electron/packager/pull/1593#discussion_r1385401573) that this file is an extra layer of protection against running Packager using unsupported Node versions. One scenario I can think of is users who added the `--ignore-engines` option to their global package manager config while working on something unrelated in the past, and then trying to run Packager and getting confused about why it doesn't work.

I'd be +1 on this if we decide that the number of users who unknowingly bypass engine constraints is sufficiently small to not be a problem.

PR #1783, opened 2024-12-17
ci: test node 22

erickzhao:
Adds CI for the latest LTS version of Node.js

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,6 +56,7 @@ workflows:
                 - node/macos
                 - node/windows
               node-version:
+                - 22.9.0
                 - 20.2.0
                 - 18.14.0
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -56,6 +56,7 @@ workflows:
                 - node/macos
                 - node/windows
               node-version:
+                - 22.9.0
```
erikian, line 59: question: do we plan to stop testing on older versions once we switch to Node 22?

erikian (approved):

.circleci/config.yml
erickzhao: Yeah, I think once we perform the big bump we'll bump `engines` everywhere

PR #1806, opened 2025-07-13
feat: support Icon Composer files as icon input for macOS

MarshallOfSound:
Validated locally but impossible to test in CI as it requires Xcode@26 beta.

This adds support for the new Icon Composer icon format in macOS 26 which is **required** in order to get liquid glass icons. Command for `actool` was reverse engineered by intercepting xcode's internal build commands and identifying the magic incantation to build icon files correctly.

In order to support both macOS >= 26 and macOS < 26 you must provide `icon: ['old.icns', 'new.icon']` as your config, pretty simple tho.


MarshallOfSound:
@dsanders11 Worth noting that so far I've only gotten this working with super specific versions of Xcode 26 beta and **only** on macOS 26 itself.

MarshallOfSound:
> Does the CLI work as expected with specifying both old and new macOS icons?

Haven't tested the CLI but assuming it converts multiple incantations to an array of strings it will work just fine. I also thing we don't need the array if we're ok requiring people to do `path/to/foo` and having `foo.icns` and `foo.icon`.

I agree an object is a good end state here, but I think that requires some more careful though, i.e. `darwinLiquidGlass` is not good naming, the `.icon` format is more generic than just lg support

PR #1830, opened 2025-09-17
refactor: improve `Options` types

erickzhao:
This PR attempts to simplify how we handle the `Options` interface in Electron Packager.

There are two main problems to address:

1. External: the `TargetArch` and `TargetPlatform` types used in the `Options` interface are aliases to `string`. This is because we accept `string` values for custom mirrors of Electron, so any string const value passed in gets swallowed by the resulting type.

  ```ts
  // the following types are functionally equivalent in TS!
  type TargetArch1 = 'arch1' | 'arch2' | string;
  type TargetArch2 = string;
  ```
2. Internal: the `Options` interface defines what options can be passed in as values, but also serves as the interface that we use to pass to the final packaging function. Along the way, many optional properties in `Options` are inferred and mutated. This approach makes our types inaccurate internally because we're using `Options` everywhere, even after inference occurs and some optional properties become required.

At a high level, we're addressing the two problems with some type refactors.

1. We no longer support arbitrary `string` values as `platform` and `arch` inputs in the types. This removes the superfluous `ArchOption`, `PlatformOption`, `TargetArch`, and `TargetPlatform` types. Users passing custom arch/platform values can still do so, but must cast their parameters using the `OfficalArch` and `OfficialPlatform` types.
2. `Options` now explicitly refers to the options that are passed in by the user. `ProcessedOptions` refers to the Options object after it is processed/inferred via metadata. `ProcessedOptionsWithSinglePlatformArch` refers to the ProcessedOptions object once it is guaranteed to have a single platform or arch value (rather than an array or comma-separated string). We use a fresh opts object every time the shape of our options changes rather than mutating the initial Options being passed in.

> [!NOTE]
> There are some additional refactors we can make for future work to make this delineation clearer.
> For example, the `Packager` class takes in `ProcessedOptions` as a parameter, which means `opts.arch` and `opts.platform` can be an array. However, these values are never used—`Packager` instance functions are only ever called with a separate `ProcessedOptionsWithSinglePlatformArch` arg.

### DIFF
--- a/src/common.ts
+++ b/src/common.ts
@@ -3,7 +3,7 @@ import fs from 'graceful-fs';
 import os from 'node:os';
 import path from 'node:path';
 import createDebug from 'debug';
-import { ComboOptions, Options } from './types.js';
+import { ProcessedOptionsWithSinglePlatformArch, Options } from './types.js';
 import { CreateOptions as AsarOptions } from '@electron/asar';
 
 export const debug = createDebug('electron-packager');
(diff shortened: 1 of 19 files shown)


MarshallOfSound (approved):
Seems fine at a glance, will be hard to know how it plays until the PR integrating into forge shows up

PR #1834, opened 2025-09-18
chore(cli)!: remove `--tmpdir=false` flag

erickzhao:
BREAKING CHANGE: removes `--tmpdir=false` option in the CLI. Use `--no-tmpdir` instead.

Removing some old deprecated CLI code. Follows the usage docs that say to use `--no-tmpdir`... but I took a look and it seems like `--no-tmpdir` isn't currently wired up in `parseArgs`? So I added it.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -20,10 +20,19 @@ async function printUsageAndExit(isError: boolean) {
 
 export function parseArgs(argv: string[]) {
   const args = yargs(argv, {
-    boolean: ['all', 'deref-symlinks', 'junk', 'overwrite', 'prune', 'quiet'],
+    boolean: [
+      'all',
+      'deref-symlinks',
+      'junk',
+      'overwrite',
+      'prune',
+      'quiet',
+      'tmpdir',
+    ],
     default: {
       junk: true,
       prune: true,
+      tmpdir: true,
     },
     string: ['electron-version', 'out'],
   });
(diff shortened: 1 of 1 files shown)


dsanders11 (approved):
Unrelated note, but we should probably take this time to drop `yargs-parser` in favor of the Node.js built-in `utils.parseArgs`.

erickzhao:
> Unrelated note, but we should probably take this time to drop yargs-parser in favor of the Node.js built-in utils.parseArgs.

Is there any breaking behaviour you envision for a refactor like that? I'm okay with saving it for a minor release of v19 if it's transparent to the user.

One other alternative we might want to consider is to lean the other way and switch to `commander` (which is what Forge uses) to align packages.

dsanders11:
> Is there any breaking behaviour you envision for a refactor like that? I'm okay with saving it for a minor release of v19 if it's transparent to the user.

I'm not familiar enough with this codebase to know for sure, but I think from a quick glance it should be possible to do without a breaking change.

> One other alternative we might want to consider is to lean the other way and switch to `commander` (which is what Forge uses) to align packages.

It looks like the CLI flags in this package are pretty simple, I don't think we need to complexity of `commander`, but it looks like `commander@^14` is zero dependency, so I'm not necessarily opposed. We can start with the built-in and switch to `commander` if complexity requires it.

PR #1867, opened 2026-01-15
fix: correct return value for `serialHooks` helper

erickzhao:
This PR rewrites `serialHooks` to actually work as intended (by passing the resulting arg as a single mega-hook into Packager's options).

Usage would look something like this:

```ts
const opts = {
  afterCopy: serialHooks([
    async () => {}, // hook 1
    async () => {}, // hook 2
    async () => {}, // hook 3
  ])
}
```

### DIFF
--- a/.github/workflows/canary.yml
+++ b/.github/workflows/canary.yml
@@ -13,13 +13,13 @@ jobs:
         os: [windows-latest, macOS-latest, ubuntu-latest]
 
     steps:
-      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # tag: v5.0.0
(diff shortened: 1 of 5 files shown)


erickzhao:
@malept I think I just wrote the `serialHooks` test wrong the first time while I was refactoring. The original problem was that I wrote `serialHooks` as an alternative to `runHooks` where it should actually be passed _into_ the latter function. I think the test as-is should be good in catching regressions?

https://github.com/electron/packager/blob/b25e5bce371961ff0b075b48e80a02c3648687b9/test/hooks.spec.ts#L45-L53

malept:
Ah OK, that makes sense.

dsanders11 (approved):

PR #1878, opened 2026-02-09
fix: ensure ASAR integrity buffer on windows is valid

### LINKED ISSUE #1873: ASAR integrity corrupt in v19
### Preflight Checklist

### Issue Details

* **Electron Packager Version:**
  * 19.0.2
* **Electron Version:**
  * 40.1.0
* **Operating System:**
  * Windows 11
* **Last Known Working Electron Packager version:**:
  * 18.4.4

### Expected Behavior
Asar integrity resource embeded into the application should be a json payload i.e
```json
[{"file":"resources\\app.asar","alg":"SHA256","value":"..."}]
```

### Actual Behavior
Integrity generated ends up being completely corrupt when viewed with Resource Hacker there is a list of files at the top, then the json payload that should be present then a bunch of corrupt data. This results in electron not launching with an error "Invalid integrity config: NOT a valid JSON.".

### To Reproduce
Package any Electron app that would have an asar file with the fuse flipped to ensure asar integrity, the packaged application will not launch on windows stating the integrity is invalid. When looking in Resource Hacker the integrity resource will be corrupt and not just a json string.

UndefinedOffset:

**Summarize your changes:**

This pull request fixes #1873 which was caused by the change in 8d64aa1777af1c714f7c9cd67aae7b132635e7fe to `resedit.js` where `.buffer` was added when setting the data for the integrity resource resulting in it being written corruptly to the final `exe`.

The documentation for [Buffer.buffer](https://nodejs.org/api/buffer.html#bufbuffer) warns the result "is not guaranteed to correspond exactly to the original Buffer". This not corresponding exactly to the original buffer seems to be what is resulting in the corrupt integrity data as reverting that change results in the correct data being written to the resource and Electron starting as expected with integrity checking enabled.

It does make me wonder if the other uses of `.buffer` could be causing issues but I did not revert the other `.buffer` as it seems to be just matching up with behavior above it.

### DIFF
--- a/src/resedit.ts
+++ b/src/resedit.ts
@@ -153,10 +153,16 @@ export async function resedit(exePath: string, options: ExeMetadata) {
       alg: options.asarIntegrity![file].algorithm,
       value: options.asarIntegrity![file].hash,
     }));
+
+    const integrityBuffer = Buffer.from(JSON.stringify(integrityList), 'utf-8');
+
     res.entries.push({
       type: 'INTEGRITY',
       id: 'ELECTRONASAR',
-      bin: Buffer.from(JSON.stringify(integrityList), 'utf-8').buffer,
+      bin: integrityBuffer.buffer.slice(
+        integrityBuffer.byteOffset,
+        integrityBuffer.byteOffset + integrityBuffer.length,
+      ),
       lang: languageInfo[0].lang,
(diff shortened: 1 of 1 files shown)


UndefinedOffset:
@MarshallOfSound should be good now :)

PR #1890, opened 2026-03-30
feat: asar integrity digest

erickzhao:
### Description

This PR implements the logic from electron/asar#380 and electron/forge#4159 directly into the Packager pipeline.

The pivot to implementing this logic in `@electron/packager` means that we get to extend the existing `AsarIntegrity` handling logic and provide a zero-config way of implementing the integrity digest for consumers. Packager and Forge users will need to just bump up the version of Packager they're consuming to get this feature.

For more details, see the original Electron PR: https://github.com/electron/electron/pull/48587.


nmggithub (approved):
Tentatively approving this. As long as it works, I think it's good. My only concern would be how difficult this may be to refactor if we ever add a new version, but we might just have to cross that bridge when we come to it.

src/mac.ts
```diff
@@ -514,6 +707,7 @@ export class MacApp extends App implements Plists {
     await this.renameElectron();
     await this.renameAppAndHelpers();
     await this.copyExtraResources();
+    await this.setIntegrityDigest();
     await this.signAppIfSpecified();
     await this.notarizeAppIfSpecified();
     return this.move();
```
mmaietta, line 801: Minor, but would it be worth extracting a helper function or a comment on the order of operations here? I see the duplicated code in `universal`.  Asking as I think it'd require contextual knowledge (or AI cross-referencing?) to know that the app can't be modified post-signing.

src/mac.ts
```diff
@@ -454,6 +571,82 @@ export class MacApp extends App implements Plists {
...
+  static INTEGRITY_DIGEST_SENTINEL = 'AGbevlPCksUGKNL8TSn7wGmJEuJsXb2A';
+
+  async setIntegrityDigest() {
+    if (!this.opts.electronVersion || !semver.valid(this.opts.electronVersion)) {
+      debug(
+        `Cannot determine Electron version (got "${this.opts.electronVersion}"), skipping integrity digest`,
+      );
+      return;
+    }
+    if (!semver.gte(this.opts.electronVersion, '41.0.0-alpha.1')) {
```
mmaietta, line 629: Minor nit: Can we add a `debug` log here to match the log pattern for "Cannot determine Electron version"?

src/mac.ts
```diff
@@ -2,17 +2,134 @@ import { App } from './platform.js';
...
+
+    const payload = Buffer.allocUnsafe(PAYLOAD_LEN);
+    payload.writeUInt8(1, 0); // used = true
+    payload.writeUInt8(1, 1); // version = 1
+    digest.copy(payload, 2); // 32-byte SHA256 digest
+
+    try {
+      for (const base of writePositions) {
+        await handle.write(payload, 0, payload.length, base);
+      }
```
mmaietta, line 129: Full disclosure: Claude is a close comrade.

----

Based on this loop, multiple sentinels would be patched, not specifically selected, when the runtime only reads one.

The runtime declares exactly one `kIntegrityDictionaryDigest` in `__DATA_CONST,__asar_integrity`, so the expected match count is 1 (or 2 for fat binaries containing both slices).

Calling this out as maybe we could also have a `sentinels.length` check to mirror the functionality/expectations/contract that `electron/fuses` [L148-L154](https://github.com/electron/fuses/blob/abf5e3d8cb7c8eb7671a2b59d38ae31adf1b0d0f/src/index.ts#L148-L154) utilizes
```
    if (sentinels.length > 2) {
      throw new Error(
        `Found ${sentinels.length} copies of the fuse sentinel in the provided Electron binary. ` +
          'At most 2 are expected (one per slice of a universal macOS binary). ' +
          'This may indicate a corrupted binary or an unsupported build configuration.',
      );
    }
```

src/mac.ts
```diff
@@ -454,6 +571,82 @@ export class MacApp extends App implements Plists {
...
+  async setIntegrityDigest() {
+    if (!this.opts.electronVersion || !semver.valid(this.opts.electronVersion)) {
+      debug(
+        `Cannot determine Electron version (got "${this.opts.electronVersion}"), skipping integrity digest`,
+      );
+      return;
+    }
+    if (!semver.gte(this.opts.electronVersion, '41.0.0-alpha.1')) {
+      return;
+    }
```
mmaietta: Interesting callout from Claude. There are multiple instances of `warning(...); return;` blocks, but that doesn't mirror the approach of `writePositions.length === 0` that throws an Error. These three preconditions warn-and-continue and also can be `--quiet`ed to be fully suppressed from dev logs.
Calling this out as there's no throw, so `signAppIfSpecified` can then sign and notarize an app whose framework slot still has `used = 0`.
```
    await this.setIntegrityDigest();
    await this.signAppIfSpecified();
    await this.notarizeAppIfSpecified();
```
According to Claude, the runtime explicitly fails open in that case (w/ // No digest to validate against, fail open in [integrity_digest.mm](https://github.com/electron/electron/blob/2d5597b1b0fa697905380184e26c9f0947e05c5d/shell/common/asar/integrity_digest.mm#L48-L49)), creating a no-op that might be difficult for devs to debug.
```
  if (kIntegrityDictionaryDigest.used == false)
    return true;  // No digest to validate against, fail open
```

nmggithub: Runtime fails open on purpose because this is an optional feature. If it fails to apply, it's up to the packager to decide if it wants to continue.

src/mac.ts
nmggithub: I much prefer an open loop. Fat Mach binaries can technically support any number of binary images. I've seen some with three: 1 x86 image and 2 ARM ones (each for slightly different CPU configurations). Even if that would never happen with us, limiting it to 2 is completely unnecessary.

mmaietta: Fair point! I guess the callout was to mirror the logic approach - should fuses restriction be relaxed then or are they completely decoupled? Maybe I misunderstood something here.
The high entropy sentinel should effectively minimize a chance of a stay collision, but if one were to happen, it'd be a fun adventure for someone to trace back.  Maybe could worth validating expected account versus slice count to avoid setting a hard limit? Or maybe just adding a debug log here to call out how many slices were identified?

nmggithub: I think the sentinel is high entropy enough to where it won't matter. I don't think the limit of 2 for the Fuses was a collision(false positive)-reduction feature, but just a simplification of the logic. If it were a collision-reduction feature, it, as a count, wouldn't even be a good one as it's would *technically* be possible for two matches in the same binary image.

I understand your paranoia, though. I'd prefer simply debug logging each time we write a digest instead of once at the end of the loop. I thought we already did the former, but perhaps not.

xgiovio:
@erickzhao Hi Erik, let us know if there are any updates on the review. thanks

georgexu99 (approved):

xgiovio:
npx asar integrity-digest on "${appBundlePath}" , when this will be ready to replace the asar command? and how is it related to https://github.com/electron/electron/pull/48587 ?
I'm bit confused right now

PR #1898, opened 2026-04-01
chore(ci): group Dependabot updates

erikian:
* [X] I have read the [contribution documentation](https://github.com/electron/packager/blob/main/CONTRIBUTING.md) for this project.
* [X] I agree to follow the [code of conduct](https://github.com/electron/electron/blob/main/CODE_OF_CONDUCT.md) that this project follows, as appropriate.
* [X] The changes are appropriately documented (if applicable).
* [X] The changes have sufficient test coverage (if applicable).
* [X] The testsuite passes successfully on my local machine (if applicable).

**Summarize your changes:**

This creates three separate [groups](https://docs.github.com/en/code-security/tutorials/secure-your-dependencies/optimizing-pr-creation-version-updates) for Dependabot updates ~~across the `github-actions` and `npm` package ecosystems~~ for the `github-actions` ecosystem so that we can get fewer PRs with multiple dependencies each, both for security fixes and for regular, scheduled version bumps:

- `major-version-updates`: every scheduled non-security major bump gets a separate PR (same as today): it's easier to fix breaking changes if there's only one variable to consider. (We might want to remove this group if we're using an older version of some package in order to reduce noise for dependencies we can't update for whatever reason)
- `non-major-version-updates`: all scheduled non-security minor/patch updates are grouped in one big PR updating them all: I think this is especially useful for the scheduled monthly updates, as it would significantly reduce the noise from multiple nearly-simultaneous PRs
- `security-updates`: all security updates (major/minor/patch, though in practice most are minor/patch) are grouped in one big PR updating them all: useful in cases where multiple packages are compromised as a result of the same attack or multiple unrelated patches come out at roughly the same time. For one-off updates, we still get a PR with only one dependency bump. Security updates are not subject to the same monthly schedule as regular dependency bumps and should still be made available right away.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,18 @@ updates:
   directory: "/"
   schedule:
     interval: "monthly"
+  groups:
+    # Creates a separate PR for each non-security-related major update.
+    major-version-updates:
+      applies-to: "version-updates"
+      group-by: "dependency-name"
+      update-types: ["major"]
+
+    # Creates a single PR with all non-security-related minor/patch updates.
+    non-major-version-updates:
+      applies-to: "version-updates"
+      update-types: ["minor", "patch"]
+
+    # Creates a single PR with all security-related updates.
+    security-updates:
+      applies-to: "security-updates"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -19,3 +19,22 @@ updates:
     # Creates a single PR with all security-related updates.
     security-updates:
       applies-to: "security-updates"
+- package-ecosystem: "npm"
+  directory: "/"
+  schedule:
+    interval: "monthly"
+  groups:
+    # Creates a separate PR for each non-security-related major update.
+    major-version-updates:
+      applies-to: "version-updates"
+      group-by: "dependency-name"
+      update-types: ["major"]
+
+    # Creates a single PR with all non-security-related minor/patch updates.
+    non-major-version-updates:
+      applies-to: "version-updates"
+      update-types: ["minor", "patch"]
+
+    # Creates a single PR with all security-related updates.
+    security-updates:
+      applies-to: "security-updates"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -19,7 +19,7 @@ updates:
     # Creates a single PR with all security-related updates.
     security-updates:
       applies-to: "security-updates"
-      update-types: ["minor", "patch"]
+      update-types: ["major", "minor", "patch"]
 - package-ecosystem: "npm"
   directory: "/"
   schedule:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -4,3 +4,39 @@ updates:
...
+      update-types: ["minor", "patch"]
+
+    # Creates a single PR with all security-related updates.
+    security-updates:
+      applies-to: "security-updates"
+      update-types: ["minor", "patch"]
+- package-ecosystem: "npm"
+  directory: "/"
+  schedule:
+    interval: "monthly"
```
erikian: note: each package ecosystem has its own grouping rules, so I had to create a new update rule for `npm`. The `schedule` here applies only to `version-updates`; the docs don't say this explicitly (why would they 🥲), but that seems to be the case since the [`schedule`](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#schedule-) key doesn't have a shield icon next to it, and as noted at the top of that same page:

> All options marked with a [shield] icon also change how Dependabot creates pull requests for security updates, except where `target-branch` is used.

If I'm reading this correctly, this means that _security_ updates are not subject to the monthly schedule and should be made available ASAP just as they are today, but they will still follow the grouping rules we're setting up here since the [`groups` key](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#groups--) does have the shield icon next to it.

.github/dependabot.yml
```diff
@@ -4,3 +4,39 @@ updates:
...
+    # Creates a separate PR for each non-security-related major update.
+    major-version-updates:
+      applies-to: "version-updates"
+      group-by: "dependency-name"
+      update-types: ["major"]
+
+    # Creates a single PR with all non-security-related minor/patch updates.
+    non-major-version-updates:
+      applies-to: "version-updates"
+      update-types: ["minor", "patch"]
```
erikian: note: if we don't want scheduled non-security version bumps for `npm` packages every month (I _think_ we currently don't have those enabled, but don't quote me on that), we can just remove those groups without affecting the grouping for security-only updates.

Since these updates would be made available in larger batches, though, it might make sense to have them enabled so we're always using relatively new versions of every package, as there would be fewer PRs to check.

dsanders11 (commented):
I think I'd prefer to only do this for the "github-actions" ecosystem at the moment, and leave the "npm" ecosystem as-is currently.

One rationale being, Dependabot is already struggling pretty hard these days doing npm updates (I have to do them manually sometimes) and will create PRs with titles that are actually listing the wrong version based on the changes in the lockfile - I think part of these struggles are due to it not playing well with the minimum age gate we've added. I'm concerned that adding groups on top of this will further exacerbate Dependabot's struggles with creating these PRs.

erickzhao (approved):

dsanders11 (approved):

PR #1906, opened 2026-04-19
fix: remove get-package-info and resolve dependencies

MarshallOfSound:
Replaces `get-package-info` with a small async generator in `infer.ts` that walks up the directory tree yielding parsed `package.json` files. `getMetadataFromPackageJSON` reads the needed fields (`productName`/`name`, `version`, `author`, `dependencies.electron`/`devDependencies.electron`/`*-nightly`) directly with optional chaining — no more dot-path-string lookup, no intermediate `result.values['dependencies.electron']` indirection, and no catch-and-reinterpret of a custom error type.

Also replaces the `resolve` package with `createRequire(src).resolve()` for locating the `electron` / `electron-nightly` `package.json` during version inference.

Drops ~35 transitive packages from the production install tree (roughly 27%), including `bluebird`, `lodash.get`, `debug@2`, `read-pkg-up`, `normalize-package-data`, `hosted-git-info`, `semver@5`, and the `spdx-*` / `validate-npm-package-license` chain.

#### Notes

- `import.meta.resolve` was not used because its two-argument `parent` form is still behind `--experimental-import-meta-resolve`; without the flag the second argument is silently ignored, which would resolve from the wrong directory. `createRequire().resolve()` is the stable equivalent.
- Node's `require.resolve` honors the `exports` field while the `resolve` package did not. `electron` and `electron-nightly` have no `exports` field, so this has no effect today.
- BOM stripping on `package.json` reads is retained for parity with the old `load-json-file` dependency.

Also removes the now-unused `src/ambient.d.ts` and the stale `get-package-info` DEBUG namespace reference from `CONTRIBUTING.md`.


erickzhao (approved):
Neat, enjoying the new control flow on this one!

PR #1917, opened 2026-06-04
fix: swap `extract-zip` with `@electron-internal/extract-zip`

erickzhao:
Does what it says on the tin.


MarshallOfSound (approved):

mkmakes:
@MarshallOfSound 

This would be great to make it into [forge](https://github.com/electron/forge/issues/4277) too! 

Thanks for everything you do!

